PR #23, opened 2017-12-26
Secondary house database update

tiusty:
Fixing the under_score in the models. Realized that I was over using the under_score so it was losing meaning, since model fields need to be accessed directly sometimes, aka admin page/model filtering. Also it was creating issues (with naming) with the forms since the forms inherit the model fields. Lastly, renamed the model classes to have model in the name for more clarify


tdefroscia (approved):
Good stuff

seanrayment (approved):

PR #24, opened 2017-12-28
Update survey models

tiusty:
Updated survey model and forms. Major overhaul. Survey is mostly tested and much cleaner. Admin page is much cleaner and it is much easier to create the initial database. Furthermore, one major change was making the houseDatabase use a foreign key instead of a charField for the home_type. Makes searching/filtering/ most things easier.


Unicorn/settings/Global_Config.py
```diff
@@ -3,7 +3,7 @@
 
 # Survey Global Variables
 default_buy_survey_name = "Recent Buy Survey"
-default_rent_survey_name = "Recent Rent Survey"
+DEFAULT_RENT_SURVEY_NAME = "Recent Rent Survey"
```
tdefroscia, line 6: Stylistically, is there any reason some globals are caps and some aren't?

survey/models.py
```diff
@@ -1,441 +1,334 @@
...
-        else:
-            type_output = ""
-            counter = 0
-            for homeType in home_type_set:
-                if counter == 0:
-                    type_output = str(homeType)
-                    counter += 1
-                else:
-                    type_output = str(homeType) + ", " + type_output
+class Destinations(models.Model):
```
tdefroscia: DestinationsModel for consistency?

tdefroscia: Perhaps more generally, what is our convention on suffixing with "model"

tdefroscia (approved):
See comments for minor style stuff (mostly just for my own clarification going forward). Other than that code looks good to me.

Unicorn/settings/Global_Config.py
tiusty: haha, no, I have been capitalizing them as I needed them. I guess I should probably just go through and capitalize all of them. Usually global definitions are written in all CAPS, which is why I am slowly switching them

survey/models.py
tiusty: yes I should switch that. My convention that I was using was, any model classes, are suffixed with Model (aka inheriting models.Model or inherit classes that inherit that). Any Form class is suffixed with Form (aka any form that inherits a formAbstractClass such as a ModelForm

tiusty: the reason why it can be important is because I am trying to keep the Class names the same for the form and the model (since they map almost directly). So the model version would be i.e RentSurveyModel and the Form version is RentSurveyForm. Therefore, if they both get imported to the same file they won't conflict

seanrayment (commented):
Should we add the listing provider as a readonly field to the admin page? That way we can see where the data came from when we have multiple providers.

tiusty:
sure! The admin pages are also very easy to edit and add stuff in the future. So we can always just add more fields if we want to later. I will add the listing provider to the admin page though as read only

tiusty:
actually not making it read only since there is no default value and it needs to be provided on new models

seanrayment (approved):

PR #36, opened 2018-01-09
Feature/modularize algorithm retrieving data

seanrayment:


houseDatabase/management/commands/pull_mlspin.py
```diff
@@ -1,19 +1,129 @@
...
+import os, sys
 from datetime import datetime
 from django.core.management.base import BaseCommand, CommandError
 from houseDatabase.models import HousePhotosModel, RentDatabaseModel, InteriorAmenitiesModel, BuildingExteriorAmenitiesModel
+from houseDatabase.management.commands.mls_fields import *
+from Unicorn.settings.Global_Config import gmaps_api_key
+from houseDatabase.models import HomeTypeModel
 from django.utils import timezone
 
+class MlspinRequester:
```
tiusty, line 14: Could we write the class comments for this class. Aka to know what it does and the data types for each member variable

houseDatabase/management/commands/pull_mlspin.py
```diff
@@ -1,19 +1,129 @@
...
+                new_listing.apartment_number_home = apartment_no
+
+                #TODO: Actually get photos based on ftp url and AWS S3
+                new_listing.save()
+                newPhotos = HousePhotosModel(house_photo=new_listing)
+                newPhotos.save()
+                new_listing.save()
+                print(full_add + " added")
+
+        print(len(lines))
```
tiusty: we should remove debugging lines from the PR?

houseDatabase/management/commands/pull_mlspin.py
```diff
@@ -1,19 +1,129 @@
...
+
 
 class Command(BaseCommand):
```
tiusty: Could we also comment this class

tdefroscia:
All tests are passing, so probably safe to merge, but Alex definitely should OK everything first.

survey/cocoon_algorithm/rent_algorithm.py
```diff
@@ -102,6 +59,72 @@ def run_compute_commute_score_exact(self):
...
+                print("Caught: " + e.__class__.__name__)
+
+        # 3: Recompute failed homes using new DB data.
+
+        for home in self.homes:
+            if len(home.approx_commute_times) < len(self.destinations):
+                # Recompute missing destinations
+                for destination in self.destinations:
+                    if destination.destination_key not in home.approx_commute_times:
+                        new_in_database = home.calculate_approx_commute(home.home.zip_code, destination, "driving")
```
tiusty: This should not be hardcoded since we want to support multiple commute types. Since rent algorithm inherits CommuteAlgorithm it should support the self.commute_type from that class

survey/cocoon_algorithm/rent_algorithm.py
```diff
@@ -102,6 +59,72 @@ def run_compute_commute_score_exact(self):
...
+        """
+
+        # 1: Query DB and update when info is there
+        failed_home_dict = dict()
+        # destination will be a DestinationsModel object
+        for destination in self.destinations:
+            failed_list = []
+            # home will be a HomeScore object
+            for home in self.homes:
+                in_database = home.calculate_approx_commute(home.home.zip_code, destination, "driving")
```
tiusty: This should also not be hardcoded per other comment

survey/cocoon_algorithm/base_algorithm.py
```diff
@@ -1,3 +1,13 @@
...
+from django.db.models import Q
+
+# Import houseDatabase modules
+from houseDatabase.models import RentDatabaseModel, HomeTypeModel
+
+# Import HomeScore class
+from survey.home_data.home_score import HomeScore
+
+
 class CocoonAlgorithm(object):
```
tiusty, line 11: The class comments should be updated (I can do this one)

survey/cocoon_algorithm/rent_algorithm.py
```diff
@@ -102,6 +59,72 @@ def run_compute_commute_score_exact(self):
...
+                # code_and_distance is a 2 element list, first an error code and second the commute time in minutes
+                if in_database != True:
+                    failed_list.append((home.home.zip_code, home.home.state))
+            # Add to the dictionary of failed homes
+            failed_home_dict[(destination.zip_code, destination.state)] = failed_list
+
+        # 2: Use approx handler to compute the failed home distances and update db
+        for destination, origin_list in failed_home_dict.items():
+            try:
+                compute_approximates.approximate_compute_handler(origin_list, destination, "driving")
```
tiusty: Should also not be just driving

survey/cocoon_algorithm/rent_algorithm.py
```diff
@@ -102,6 +59,72 @@ def run_compute_commute_score_exact(self):
...
+
+        for destination in self.destinations:
+            try:
+                # map list of HomeScore objects to full addresses
+                origin_addresses = list(map(lambda house:house.home.full_address,
+                                       self.homes[:number_of_exact_commutes_computed]))
+
+                destination_address = destination.full_address
+
+                results = distance_matrix_requester.calculate_distances(origin_addresses, [destination_address])
```
tiusty: Shouldn't this take in a commute type?

survey/cocoon_algorithm/rent_algorithm.py
```diff
@@ -102,6 +59,72 @@ def run_compute_commute_score_exact(self):
...
+        # 2: Use approx handler to compute the failed home distances and update db
+        for destination, origin_list in failed_home_dict.items():
+            try:
+                compute_approximates.approximate_compute_handler(origin_list, destination, "driving")
+            except Distance_Matrix_Exception as e:
+                print("Caught: " + e.__class__.__name__)
+
+        # 3: Recompute failed homes using new DB data.
+
+        for home in self.homes:
```
tiusty, line 103: I would prefer not to call the variable home since it isn't a home.
Before I did:
for home_data in self.homes to differentiate from the home stored in the homeScore
but maybe this would be better:
for home_score in self.homes:

tiusty: I have done this too and I should probably fix mine as well

survey/cocoon_algorithm/rent_algorithm.py
```diff
@@ -102,6 +59,72 @@ def run_compute_commute_score_exact(self):
...
+            try:
+                # map list of HomeScore objects to full addresses
+                origin_addresses = list(map(lambda house:house.home.full_address,
+                                       self.homes[:number_of_exact_commutes_computed]))
+
+                destination_address = destination.full_address
+
+                results = distance_matrix_requester.calculate_distances(origin_addresses, [destination_address])
+
+                for i in range(min(len(self.homes[:number_of_exact_commutes_computed]), len(results))):
```
tiusty: Can you comment on what this is doing? I think I know but a lot of things are happening

survey/distance_matrix/compute_approximates.py
```diff
@@ -0,0 +1,60 @@
+from houseDatabase.models import RentDatabaseModel, HomeTypeModel, ZipCodeDictionaryParentModel, ZipCodeDictionaryChildModel
+from survey.distance_matrix.distance_wrapper import DistanceWrapper
+
+# TODO: best way to add more specificity than zip code
+# TODO: update values after certain length of time
+def approximate_compute_handler(origins_zips_states, destination_zip_state, commute_type):
```
tiusty: Is it supposed to be, compute or commute? Maybe a better name would be calculate_approximate_commute

survey/distance_matrix/compute_approximates.py
```diff
@@ -0,0 +1,60 @@
...
+        else:
+            ZipCodeDictionaryParentModel.objects.create(zip_code_parent=origin[0]) \
+                .zipcodedictionarychildmodel_set.create(
+                zip_code_child=destination_zip_state[0],
+                commute_type_child=commute_type,
+                commute_distance_meters_child=result[0][1],
+                commute_time_seconds_child=result[0][0],
+            )
+
+    return results
```
tiusty: why do we need a return value? I feel like a boolean saying if it succeeded for failed is fine.

tiusty: It looked like the return isn't even being used in the code

survey/distance_matrix/distance_wrapper.py
```diff
@@ -0,0 +1,143 @@
+from googlemaps import distance_matrix, client
+from Cocoon.settings.Global_Config import gmaps_api_key
+
+class DistanceWrapper:
+
```
tiusty: Class comments would be good

tiusty: i.e 
    """
    Class included the base values for any algorithm.

      Attributes:
        self._homes (List[HomeScore]): Stores a list of homes as a HomeScore
        self._destinations (List[anything that inherits survey.models.DestinationsModel]): The desired destinations
            specified by the user. These locations are the work, schools that the user needs to go to.

    """

houseDatabase/management/commands/pull_mlspin.py
tdefroscia: Yes

survey/cocoon_algorithm/rent_algorithm.py
tdefroscia: This is definitely the correct approach, will change.

survey/cocoon_algorithm/rent_algorithm.py
tdefroscia: Sean will have to take a look at this

survey/cocoon_algorithm/rent_algorithm.py
tdefroscia: Also for Sean to look at

survey/distance_matrix/distance_wrapper.py
```diff
@@ -0,0 +1,143 @@
...
+            origins: ["2 Snow Hill Lane, Medfield MA"]
+            destinations: ["Boston, MA"]
+            destinations: ["23412", "159 Brattle Street, Arlington MA"]
+        """
+
+        distance_matrix_list = []
+
+        origin_list = origins
+        while origin_list:
+            if (len(origin_list) > 25):
```
tiusty: This will fail if there is more than one destination. The distance matrix api can only take 25 elements. An element is defined by (origins * destinations) = element. So the amount the list should be spliced by should be variable depending on the number of destinations. So if there are 2 destinations than 25/2 origins should also be included
@seanrayment

survey/distance_matrix/distance_wrapper.py
```diff
@@ -0,0 +1,143 @@
...
+                response_json = distance_matrix.distance_matrix(self.client,
```
tiusty: This should also take in commute type
@seanrayment

survey/distance_matrix/distance_wrapper.py
tiusty: I also feel like having the two cases is redundant. Can't you just use the first case for all of it? because origin_list = origin_list[25:] will go to empty if there is less than 25 elements. Also for the origin size you could do something list origing_list[:25//len(destinations)] and for destinations just destinations[:25]. We should think of the best solution but it might be best to make that one case
@seanrayment

tiusty (changes requested):
Distance matrix comments

survey/home_data/home_score.py
```diff
@@ -72,6 +73,39 @@ def exact_commute_times(self, new_exact_commute_time):
         """
         self._exact_commute_times_minutes.update(new_exact_commute_time)
 
+    # TODO: Move to rent_algorithm (?)
+    # This should probably be moved into the rent_algorithm file as a helper method, as it doesn't
+    # actually interact with the object in any meaningful way.
+    def calculate_approx_commute(self, origin_zip, destination, commute_type):
```
tiusty: I kind of don't like the name of this function anymore. It doesn't really calculate the approx commute, it more checks to see if the zip code is cached. Maybe change the name to like, populate_approx_commutes or something like that.
  @tdefroscia

survey/home_data/home_score.py
```diff
@@ -72,6 +73,39 @@ def exact_commute_times(self, new_exact_commute_time):
...
+        """
+        Computes an approximate commute time for this house to an input destination. First checks
+        the zipcode database to see if the commute time is already stored; if it's not, it then
+        returns the pair of failed zips, along with an error code as a 3 element list. The first
+        entry of the list is 0 if the pair was in the database, and 1 if the pair wasn't in the database
+        or if the pair wasn't valid. The last 2 entries are the origin and destination zip respectively.
```
tiusty: Also update this documentation since you got rid of the return values
  @tdefroscia

tiusty (changes requested):
More minor things

tiusty:
The PR looks good! I just commented what I think should be changed but let me know if you think otherwise! Once we talk/resolve those issues we should be good to go!

survey/cocoon_algorithm/rent_algorithm.py
tiusty: @seanrayment

survey/cocoon_algorithm/rent_algorithm.py
tiusty: @seanrayment

survey/home_data/home_score.py
tiusty: also I think it can stay in HomeScore because it populates the variable in HomeScore but let me know if you think otherwise

survey/cocoon_algorithm/rent_algorithm.py
```diff
@@ -96,7 +104,7 @@ def retrieve_all_approximate_commutes(self):
                 # Recompute missing destinations
                 for destination in self.destinations:
                     if destination.destination_key not in home.approx_commute_times:
-                        new_in_database = home.calculate_approx_commute(home.home.zip_code, destination, self.commute_type)
+                        new_in_database = home.populate_approx_commutes(home.home.zip_code, destination, self.commute_type)
                         if new_in_database != True:
```
tiusty: I would change this to not in_database like you did earlier
@tdefroscia

tiusty (changes requested):
Minor change requested

survey/cocoon_algorithm/rent_algorithm.py
seanrayment: The DistanceWrapper obj defaults to "driving" but I will update this to make it clearer

survey/cocoon_algorithm/rent_algorithm.py
seanrayment: Thanks for taking note of this one, it was unnecessarily complex. Fixed!

survey/distance_matrix/compute_approximates.py
seanrayment: commute! thank you

survey/distance_matrix/compute_approximates.py
seanrayment: I guess the return type isn't necessary because it catches an exception on failure.

survey/distance_matrix/distance_wrapper.py
seanrayment: Added

survey/cocoon_algorithm/rent_algorithm.py
tiusty: ah I see, in the constructor of DistanceWrapper it sets it to driving.Then the constructor should take in mode-self.commute_type so it can support another commute type. I am not a big fan of that because it is confusing, Since calculate_distance (which should be renamed because it also gets time), is the only function that uses commute type, maybe it should be passed in instead and removed as a member variable?
@seanrayment

tiusty (changes requested):

survey/cocoon_algorithm/rent_algorithm.py
```diff
@@ -124,7 +125,9 @@ def retrieve_exact_commutes(self):
 
                 destination_address = destination.full_address
 
-                results = distance_matrix_requester.calculate_distances(origin_addresses, [destination_address])
+                results = distance_matrix_requester.get_durations_and_distances(origin_addresses,
+                                                                                [destination_address],
+                                                                                mode="driving")
```
tiusty: is there any reason why we can't do mode=self.commute_type? Right now the rent_algorithm itself defaults to driving. That we the rent algorithm can work with other commute types

survey/distance_matrix/compute_approximates.py
```diff
@@ -22,9 +25,10 @@ def approximate_commute_handler(origins_zips_states, destination_zip_state, comm
 
     wrapper = DistanceWrapper(mode=commute_type)
 
-    # map zip, state tuples list to a list of "zip state" strings
-    results = wrapper.calculate_distances(list(map(lambda x:x[0]+" "+x[1], origins_zips_states)),
-                                          [destination_zip_state[0]+" "+destination_zip_state[1]])
+    # map (zip, state) tuples list to a list of "zip state" strings
+    results = wrapper.get_durations_and_distances(list(map(lambda x:x[0]+" "+x[1], origins_zips_states)),
+                                                  [destination_zip_state[0]+" "+destination_zip_state[1]],
+                                                  mode="driving")
```
tiusty: same with here, since the commute_type is passed in could mode=commute_type? and then the mode can be removed from the constructor. Sorry if you are gonna do that anyways.

tiusty (changes requested):

survey/cocoon_algorithm/rent_algorithm.py
seanrayment: The reason I didn't do that is because I thought our commute_type was "Driving" and the googlemaps package uses "driving." Although I suppose it would be simple enough to change our model

tiusty: it actually supports driving, walking, biking and transit. Right now driving is the only tested one. I would like to get the other ones tested and reliable soon enough. The model field itself is a choice field with all 4 options

seanrayment: I'm referring to the "D" vs. "d". googlemaps expects lower case strings, but looking again I think our models are lowercase as well, so I can change that

tiusty: yea the model is receiving from global config: 

COMMUTE_TYPES = (
    ('driving', 'Driving'),
    ('transit', 'Transit'),
    ('walking', 'Walking'),
    ('biking', 'Biking'),
)

So i think it will display it with D, but then actually save it with d. Let me know if that is wrong.

survey/distance_matrix/distance_wrapper.py
seanrayment: Looking back over the distance matrix documentation. A request can have 100 elements and a maximum of 25 destinations or origins. This would support up to 4 destinations. Shouldn't that be enough?

tiusty: yea... you are right it is 100 elements, then what I was saying later is wrong.

survey/cocoon_algorithm/rent_algorithm.py
```diff
@@ -102,6 +68,75 @@ def run_compute_commute_score_exact(self):
...
+                                       self.homes[:number_of_exact_commutes_computed]))
+
+                destination_address = destination.full_address
+
+                results = distance_matrix_requester.get_durations_and_distances(origin_addresses,
+                                                                                [destination_address],
+                                                                                mode=self.commute_type)
+
+                # iterates over min of number to be computed and length of results in case lens don't match
+                for i in range(min(number_of_exact_commutes_computed), len(results)):
```
tiusty: Unit tests were failing due to a bug on this line. The parentheses are messed up. The min should encompass both of the elements. Good thing for unit tests! I could change it but since it your code I will let you do it to prevent confusion.
@seanrayment

tiusty (changes requested):

tiusty (approved):

PR #42, opened 2018-01-12
Update settings.py

tiusty:
Update the settings file to have both a production and a local version. The local version is meant for a local dev environment while the production is meant for the server. This will be important going forward so the split should be made now. Therefore the production machine can use a remote database but a local machine can still use the sqlite.


seanrayment (approved):
Looks good, but where is it determined which of these settings files will be active?

PR #44, opened 2018-01-14
Frontend fixes

seanrayment:


tdefroscia (approved):

survey/templates/survey/surveyResultRent.html
```diff
@@ -284,21 +284,19 @@ <h1 class="page-header">Home List</h1>
 
             </script>
 
-            {% for home in houseList %}
+            {% for house in houseList %}
```
tiusty: could we name this like home_score instead of house, will make it easier to understand

tiusty (changes requested):
Minor change

tiusty (approved):

PR #46, opened 2018-01-14
Feature update commute type

tiusty:
This update makes the commute type a foreign key instead of a char. This is much cleaner implementation and always easy addition of new commute types and linking surveys etc to commute types.


tdefroscia (approved):

tdefroscia:
This makes a lot more sense, you're right Alex

seanrayment (approved):

PR #81, opened 2018-02-08
Mls fix

seanrayment:
I think the best way to go about it would be to leave move_in_day in the model for use later on. Then last_updated and is_available work together. 

Maybe in the pre-survey static filtering we designate all apartments where the last_updated does not match the last_updated_database to is_available=false. Then we exclude all homes that are not available.


houseDatabase/models.py
```diff
@@ -283,6 +288,13 @@ def image_path(self):
         return self.image_path_photo
 
 
+class DatabaseManagementModel(models.Model):
```
tiusty: I think we should start creating database managers per apartment feed. So instead of being a generic database manager, it can be just for the MLS feed. So I would call this MLSManagementModel or something like that. Then if we get other feeds then they can have separate managers depending on how the data is fed etc

houseDatabase/management/commands/pull_mlspin.py
```diff
@@ -96,40 +98,50 @@ def parse_idx_feed(self):
...
-                new_listing.save()
                 newPhotos = HousePhotosModel(house_photo=new_listing)
                 newPhotos.save()
                 new_listing.save()
-                print(full_add + " added")
+                print("[ ADDING   ]" + full_add)
+        
+        # TODO: Assess problem of time difference between start/end of update
+        manager = DatabaseManagementModel.objects.get(pk=1)
+        manager.last_updated_database = timezone.now()
```
tiusty: to be consistent so there isn't weird issues if the script is called before midnight and then finishes after midnight. I think you should get the time when the script is called and then use that time for when all the homes are updated and when the homes are last updated (aka the databaseManagementModel thing)

houseDatabase/management/commands/pull_mlspin.py
```diff
@@ -96,40 +98,50 @@ def parse_idx_feed(self):
                     lat = latlng[0]
                     lng = latlng[1]
 
-                new_listing = RentDatabaseModel(home_type_home=(HomeTypeModel.objects.get(home_type_survey="Apartment")))
-                new_listing.currently_available = True
+                new_listing = RentDatabaseModel()
+                list_type = cells[PROP_TYPE]
+                if (list_type == "RN"):
```
tiusty: what does RN mean? Could you comment what is happening a little more? That would be super helpfull so I don't have to figure it out! :)

tiusty (changes requested):
Overall looks good but think a few changes would be benificial

tiusty:
Also commenting on your previous comment. I would prefer if the move_in_day stuff is removed since it is kinda dead code right now. It is easy enough to implement so if later we need it, we can easily add it and reimplement it the correct way based on the data feed. You could also make the argument that is_available is useless right now too.... so I could remove that if you think it has no point.

houseDatabase/management/commands/pull_mlspin.py
```diff
@@ -96,40 +98,50 @@ def parse_idx_feed(self):
...
-                    print("listing not a rental")
+                
+                try:
+                    # assert that fields are ints
+                    new_listing.price_home = int(cells[LIST_PRICE])
+                    new_listing.num_bedrooms_home = int(cells[NO_BEDROOMS])
+                    no_baths = int(cells[NO_FULL_BATHS]) + int(cells[NO_HALF_BATHS])
+                    new_listing.num_bathrooms_home = no_baths
+                    new_listing.listing_number_home = int(cells[LIST_NO])
+                except ValueError:
```
tiusty, line 126: If the fields aren't ints should the home be immediately disqualified? What is the downside of doing nothing?

houseDatabase/management/commands/pull_mlspin.py
```diff
@@ -96,40 +98,50 @@ def parse_idx_feed(self):
...
                 # TODO: upload ftp path to S3 storage bucket
-                new_listing.save()
                 newPhotos = HousePhotosModel(house_photo=new_listing)
                 newPhotos.save()
                 new_listing.save()
-                print(full_add + " added")
+                print("[ ADDING   ]" + full_add)
+        
+        # TODO: Assess problem of time difference between start/end of update
+        manager = DatabaseManagementModel.objects.get(pk=1)
```
tiusty: I don't like this method of retrieving the object cus for some reason, if it gets deleted and then gets recreated, then it won't be pk=1, it would be pk=2. Maybe we can think of a better way of retrieving the object? Maybe just DatabaseManagementMode.objects.all().first() or something. Up to you. Hopefully, we can assume that there would only be one object. You can look at this post to see how to override the save() method so only one object of the model can ever exist (in essence making it a singleton) https://stackoverflow.com/questions/39412968/allow-only-one-instance-of-a-model-in-django

tiusty: I can also work on it if you don't want to do it

tiusty:
side note.... so nice not to have  PR with 30k lines changes haha

houseDatabase/models.py
seanrayment: Agreed

houseDatabase/management/commands/pull_mlspin.py
seanrayment: Yup, good call

houseDatabase/management/commands/pull_mlspin.py
seanrayment: I'll clarify in a comment, but "RN" is the plaintext in the MLS feed that denotes the listing is a rental.

houseDatabase/management/commands/pull_mlspin.py
seanrayment: This is to protect against typos in the MLS feed, which is currently giving us a ValueError in the live version. So if the number_bathrooms field contains "c" in plaintext it triggers the ValueError and we skip the home. 

Let me know if that's not what you were asking about.

tiusty: yea I missed read it, looks good!

tiusty (approved):
Looks good! Quick questions. Was this working for you on your local machine? When I took your branch it wasn't running because of some missing imports etc. I ended up committing them.

PR #84, opened 2018-02-19
Issue78 multiple commutes

tiusty:
Allows multiple commutes to be requested on the survey


tiusty:
Also, this PR sets the default commute type to driving

seanrayment (approved):

PR #86, opened 2018-02-19
Feature/generate list

seanrayment:


survey/templates/survey/surveyResultRent.html
```diff
@@ -359,6 +416,20 @@ <h1 class="page-header">Home List</h1>
 
   <script src="{% static 'survey/markerclusterer.js' %}"></script>
 
+  <script type="text/javascript">
```
tiusty: Just to reduce the amount of javascript in the html file, could you move this to a javascript file?

tiusty (approved):
Looks good

PR #94, opened 2018-06-18
Update google api keys to use the devteam account google console

tiusty:
Approximate commutes was broken because google updated the api call


tiusty:
This PR, updates the google api key. It also fixes the issue where the approximate commute would not update. This also fixes the tests for the distance wrapper so it doesn't use the api calls and instead uses a mock function

PR #99, opened 2018-07-10
Update file structure

tiusty:
This still needs to be more thoroughly tested but this updates the file structure to the correct format. This is following the guidelines from the book I am reading


tiusty:
This has been working well for me, so it should be safe to merge. If any weird issues occur, then it might be caused by this

PR #100, opened 2018-07-11
Add images

tiusty:
This adds images being pulled from MLS locally. Some work will need to be done to get it to work with the S3 bucket but this should be good enough for now


cocoon/static/Unicorn/homes/css/tiles.css
```diff
@@ -240,10 +240,11 @@ STYLING NEW TILE RESULTS
 
 }
 .thumbnailItem {
-    /*padding: 4px;*/
+    padding: 4px;
+    display: flex;
```
tiusty, line 244: @seanrayment Can you just make sure I didn't do something stupid in the frontend

tiusty: This isn't working in chrome. I don't really understand flex right now... but I think I just made a nested flex container which is causing issues. What I want to happen is have the img fill the thumbnailItem div.

tiusty:
Seems to work with Firefox. Need to investigate into why Chrome and Safari have difficutly displaying the little tile images

PR #104, opened 2018-07-26
Issue 78 fix autocomplete

tiusty:
Using google autocomplete instead of smarty streets. 
Now instead of typing in the different fields, an address field is provided which will populate the existing fields


tiusty:
Not mergined in favor of other PR

PR #106, opened 2018-07-27
Issue 78 just autocomplete changes

tiusty:
The last branch contained undesired changes. Resetting branch to the correct spot


tiusty:
Think everything looks good

PR #130, opened 2018-08-29
Issue 127 fix biking walking approximations

tiusty:
This adds the new biking and walking approximation method. It also reworks the approximation methods to be cleaner and in the right spot.


cocoon/survey/home_data/home_score.py
```diff
@@ -101,20 +107,72 @@ def exact_commute_times(self, new_exact_commute_time):
...
+        # Stores the lat and lng points for the home
+        lat_lng_home = (home.latitude, home.longitude)
+
+        # Returns the distance from the two lat lng points in miles
+        distance = geopy.distance.geodesic(lat_lng_home, lat_lng_dest).miles
+
+        # If the distance is less than a mile then don't add any distance since it is already so close
+        if distance > 1:
+            # Extra distance is determined by giving more distance to homes farther away
+            extra_distance = EXTRA_DISTANCE_LAT_LNG_APPROX * (1 - 1.0/math.sqrt(distance))
```
tiusty, line 149: This is wrong. This should be x^2/3 because we want homes closer to have less of an effect, and have the effect to be greater the farther away the home is.

PR #189, opened 2018-09-26
[issue-184] Fixed method of reading in the idx_feed. Also it fixes bu…

tiusty:
Fixes two issues
1. idx_feed sometimes had carriage return and sometimes didn't... so fixed reading line method
2. Last home in MLSPIN not being added


tiusty:
Need to add condition where if there is a null value exception then don't save the home

akashkumar2296 (approved):

tiusty:
Fixes issues #190 and #188

PR #191, opened 2018-09-28
Itinerary model

akashkumar2296:
Most of the work done is in models.py

### DIFF
--- a/cocoon/itinerary/models.py
+++ b/cocoon/itinerary/models.py
@@ -1,0 +1,24 @@
+from django.db import models
+
+# Import cocoon models
+from cocoon.userAuth.models import MyUser
+from cocoon.houseDatabase.models import RentDatabaseModel
+
+class Itinerary(models.Model):
+	"""
+   	Model for Itinerary. These are based on the interface designed on the Google Doc. 
+
+   	NOTE: 
+   	1. Itierary File - I wasn't sure what to make FileField, so I made it null for now
+   	2. Available_Start_time - I wasn't sure how to make a list,so I just set it as a date for now
+   	3. Homes - Alex was relunctant on making this a ManyToMany
+    """
+    user = models.ForeignKey('MyUser', on_delete=models.CASCADE)
+    itinerary_file = models.FileField(Field.null, on_delete=models.CASCADE)
+    agent = models.ForeignKey('MyUser', on_delete=models.CASCADE)
+    start_time = models.DateTime(default=timezone.now)
+    available_start_time = models.DateField(default=timezone.now)
+    homes = models.ManyToMAny(RentDatabaseModel)
+
+    def __str__(self):
+        return self.itinerary_file
--- a/cocoon/itinerary/models.py
+++ b/cocoon/itinerary/models.py
@@ -14,11 +14,24 @@ class Itinerary(models.Model):
    	3. Homes - Alex was relunctant on making this a ManyToMany
     """
     user = models.ForeignKey('MyUser', on_delete=models.CASCADE)
-    itinerary_file = models.FileField(Field.null, on_delete=models.CASCADE)
+    itinerary_file = models.FileField(blank=True)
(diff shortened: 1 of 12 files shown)


cocoon/itinerary/models.py
```diff
@@ -0,0 +1,24 @@
...
+	"""
+   	Model for Itinerary. These are based on the interface designed on the Google Doc. 
+
+   	NOTE: 
+   	1. Itierary File - I wasn't sure what to make FileField, so I made it null for now
+   	2. Available_Start_time - I wasn't sure how to make a list,so I just set it as a date for now
+   	3. Homes - Alex was relunctant on making this a ManyToMany
+    """
+    user = models.ForeignKey('MyUser', on_delete=models.CASCADE)
+    itinerary_file = models.FileField(Field.null, on_delete=models.CASCADE)
```
tiusty: Reading my best practices book, it is a bad idea to set a file field to null, so setting blank=true is much better

cocoon/itinerary/models.py
```diff
@@ -0,0 +1,24 @@
...
+class Itinerary(models.Model):
+	"""
+   	Model for Itinerary. These are based on the interface designed on the Google Doc. 
+
+   	NOTE: 
+   	1. Itierary File - I wasn't sure what to make FileField, so I made it null for now
+   	2. Available_Start_time - I wasn't sure how to make a list,so I just set it as a date for now
+   	3. Homes - Alex was relunctant on making this a ManyToMany
+    """
+    user = models.ForeignKey('MyUser', on_delete=models.CASCADE)
```
tiusty, line 16: I believe this is the right way because a user can have multiple itineraries but an itinerary will never have multiple users

cocoon/itinerary/models.py
tiusty: Not sure if we want models.CASCADE for on_delete

cocoon/itinerary/models.py
```diff
@@ -0,0 +1,24 @@
...
+    itinerary_file = models.FileField(Field.null, on_delete=models.CASCADE)
+    agent = models.ForeignKey('MyUser', on_delete=models.CASCADE)
+    start_time = models.DateTime(default=timezone.now)
+    available_start_time = models.DateField(default=timezone.now)
```
tiusty: creating lists in a database is annoying.. we might need to do some sort of foreign key relationship to avaiable start times and then we can store multiple start times that way.

cocoon/itinerary/models.py
```diff
@@ -0,0 +1,24 @@
...
+
+   	NOTE: 
+   	1. Itierary File - I wasn't sure what to make FileField, so I made it null for now
+   	2. Available_Start_time - I wasn't sure how to make a list,so I just set it as a date for now
+   	3. Homes - Alex was relunctant on making this a ManyToMany
+    """
+    user = models.ForeignKey('MyUser', on_delete=models.CASCADE)
+    itinerary_file = models.FileField(Field.null, on_delete=models.CASCADE)
+    agent = models.ForeignKey('MyUser', on_delete=models.CASCADE)
+    start_time = models.DateTime(default=timezone.now)
```
tiusty: just to make it more clear I think we should call it selected_start_time.

Also, if we make the avaiable_start_time a reverse foreign key to avaiable_start_times model then we can do a 1-1 relationship to the time selected in the avaiable_start_times field

tiusty (commented):
Some comments regarding the new model

cocoon/itinerary/models.py
akashkumar2296: Ok for this, I made them both Foreign keys to point to a StartTime model and a EndTime model. The StartTime makes sense, but EndTime is a little weird cuz im trying to get the idea of a reverse foreign key, but I figured there would be more than one end time. Feel free to change it up as you guys would like.

tiusty: I don't think we need to do an end time right? Cus we will know the estimated time and the start time so we can just create a function to calculate the end time

PR #194, opened 2018-09-30
Add react frontend base

tiusty:
This PR contains the base commit for adding react to our website. Right now the react is namespaced so the original website is not affected. Also I added the base routes for all the current routes but there is no functionality right now. Aka, there is /react/survey, /react/homePage etc defined in the componenet/Main.js. Each componenet is then loaded from there. An example code of react interfacing with DRF is in react/signatures. To utilize the example you need to make sure to makemigrations and migrate. then you can go to /signatures/snippets and post some data to have data to display. Then once a few snippets are made, you can go to /react/signatures and those snippets will be displayed via the DRF. You guys can decide if you want to merge this in so you can create the scheduler via react. All you would need to do is define a react component for the scheduler and import it into the Main.js in the component directory.

The way to use react.
You need to run npm run dev to compile it then it should render when you runserver


tiusty:
Make sure to run the requirments.txt file so install the new requirments

PR #197, opened 2018-10-03
Fix geolcate bug

tiusty:
Lets try to get this in by tonight.

This fixes 2 things
1. An issue where the update would then set the lat and long to 0 (since it is not recomputed for mlspin for each new listing)
2. The itinerary model had some issues

### DIFF
--- a/cocoon/scheduler/models.py
+++ b/cocoon/scheduler/models.py
@@ -27,12 +27,13 @@ class ItineraryModel(models.Model):
            self.available_start_times (ForeignKey) -> Stores the avaiable times that the user can go on a tour
            self.homes (ManytoManyField) -> Stores the homes that are associated with this itinerary
     """
-    client = models.ForeignKey(MyUser, related_name='client', on_delete=models.CASCADE)
+    client = models.ForeignKey(MyUser, related_name='my_tours', on_delete=models.CASCADE)
     itinerary = models.FileField(blank=True)
-    agent = models.ForeignKey(MyUser, related_name='agent', on_delete=models.CASCADE, blank=True)
+    agent = models.ForeignKey(MyUser, related_name='scheduled_tours', on_delete=models.CASCADE, blank=True, null=True)
     tour_duration_seconds = models.IntegerField(default=0)
-    selected_start_time = models.OneToOneField(TimeModel, on_delete=models.CASCADE, blank=True)
-    available_start_times = models.ForeignKey(TimeModel, related_name="start_times", on_delete=models.CASCADE, blank=True)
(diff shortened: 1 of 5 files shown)


cocoon/scheduler/models.py
```diff
@@ -20,12 +27,12 @@ class ItineraryModel(models.Model):
...
-    client = models.ForeignKey('MyUser', on_delete=models.CASCADE)
+    client = models.ForeignKey(MyUser, related_name='client', on_delete=models.CASCADE)
     itinerary = models.FileField(blank=True)
-    agent = models.ForeignKey('MyUser', on_delete=models.CASCADE, blank=True)
+    agent = models.ForeignKey(MyUser, related_name='agent', on_delete=models.CASCADE, blank=True)
     tour_duration_seconds = models.IntegerField(default=0)
-    selected_start_time = models.OneToOneField('TimeModel', on_delete=models.CASCADE, blank=True)
-    available_start_times = models.ForeignKey('TimeModel', on_delete=models.CASCADE, blank=True)
+    selected_start_time = models.OneToOneField(TimeModel, on_delete=models.CASCADE, blank=True)
+    available_start_times = models.ForeignKey(TimeModel, related_name="start_times", on_delete=models.CASCADE, blank=True)
```
seanrayment: I'm totally nitpicking, but want to put this in single quotes to keep it consistent?

seanrayment (approved):
Sweet!

cocoon/scheduler/models.py
tiusty: yea I should!

cocoon/scheduler/models.py
```diff
@@ -20,12 +27,12 @@ class ItineraryModel(models.Model):
            self.available_start_times (ForeignKey) -> Stores the avaiable times that the user can go on a tour
            self.homes (ManytoManyField) -> Stores the homes that are associated with this itinerary
     """
-    client = models.ForeignKey('MyUser', on_delete=models.CASCADE)
+    client = models.ForeignKey(MyUser, related_name='client', on_delete=models.CASCADE)
     itinerary = models.FileField(blank=True)
-    agent = models.ForeignKey('MyUser', on_delete=models.CASCADE, blank=True)
+    agent = models.ForeignKey(MyUser, related_name='agent', on_delete=models.CASCADE, blank=True)
```
tiusty: I realized that I think the related_name is wrong.

I believe for agents it would be:
MyUser.agent.all() -> Which is wrong because the reverse field should be
MyUser.client_tours

so I think I should change the related_name for agent to "client_tours"

for the client field maybe the cilent field related name should be:
MyUser.itineraries.all()

i.e change the related name to itineraries

PR #199, opened 2018-10-11
Issue 181 best transit mode

akashkumar2296:
Currently the functionality allows for users to pick both transit modes (bus, subway, train) and also commute types (driving, walking, biking). I think these should both be available because you may not always want to drive/walk/bike but at the same time you might want flexibility in your commute choices. By having the transit modes appear only after "transit" is selected in commute type, you're narrowing down the list of valid homes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 .idea/
 
 # Ignore the settings file
-config/settings/secrets.json
+config/settings/secret.json
 
 # Ignore the other apache files that isn't the template file
 config/apache2/*.conf
--- a/cocoon/commutes/admin.py
+++ b/cocoon/commutes/admin.py
@@ -35,7 +35,6 @@ class TransitTypeAdmin(admin.ModelAdmin):
     ]
     list_display = ('transit_type',)
 
-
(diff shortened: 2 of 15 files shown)


tiusty:
Right now I feel like realistically we should only support one commute type. Also the user will have a main type of commuting. If we allow the user to select a transit mode and also driving/biking/walking then why don't we just let them select driving and or biking, and or walking and or transit type etc. Therefore, I think we should let them select driving/walking/biking/transit and then if they select transit then ask them the type of transits they would want. We can talk more about it tomorrow

tiusty:
Based on the google docs for distance matrix api:

transit_mode — Specifies one or more preferred modes of transit. This parameter may only be specified for requests where the mode is transit.

I disagree that we should let users choose Driving then also specify transit modes. I still believe we should have it so they have to choose Driving/Walking/Transit/Biking and then if they choose transit they can select transit options. Later we will also add driving modes when the user selects driving.

tiusty:
Declining for now because:

1. We discussed that the functionality will be changed (i.e only show transit modes when transit is selected)
2. Formatting of comments etc, and having lines of code that are changed that shouldn't be changed needs to be removed etc.

PR #203, opened 2018-10-17
Moved time foreign key to TimeModel

seanrayment:
Note: I'm not quite sure the best way to keep track of the selected tour time. The way I've implemented it we will need to be careful that only one time has been selected.

### DIFF
--- a/cocoon/scheduler/models.py
+++ b/cocoon/scheduler/models.py
@@ -5,14 +5,6 @@ from django.utils import timezone
 from cocoon.userAuth.models import MyUser
 from cocoon.houseDatabase.models import RentDatabaseModel
 
-
-class TimeModel(models.Model):
-    time = models.DateTimeField(default=timezone.now)
-
-    def __str__(self):
-        return self.time
-
-
 class ItineraryModel(models.Model):
     """
        Model for Itinerary. These are based on the interface designed on the Google Doc.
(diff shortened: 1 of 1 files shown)


tiusty (changes requested):
I would say leave the selected time in the Itinerary model. That way when a time is selected the one to one relationship is created to that time model

tiusty:
All but one of the time models will be selected, so I don't think the time models should store which time the tour is happening. The itinerary should

seanrayment:
@tiusty re-tagging now that I've made that change

cocoon/scheduler/models.py
```diff
@@ -53,3 +43,11 @@ def tour_duration_hours(self):
         :return:
         """
         return self.tour_duration_minutes/60
+
+
+class TimeModel(models.Model):
+    time = models.DateTimeField(default=timezone.now)
+    itinerary_model = models.ForeignKey(ItineraryModel, related_name='itinerary_model', on_delete=models.CASCADE)
```
tiusty: btw, the related name is the reverse look up name that is created so we want to make it intuitive for what that relationship is for.

Therefore having the related name as like availabe__start_times would be better so you can do something like:

ItineraryModel.avaiable_start_times.all() 

Right now it would be
ItineraryModel.itinerary_model.all() which is weird and less intuitive

tiusty (changes requested):

tiusty (approved):

PR #205, opened 2018-10-18
Issue 139 add docusign

tiusty:
This adds the functionality for signing documents. Specifically, this only adds the pre_tour forms. Though this should be easily expandable to more types of forms in the future.

Some notes:
1. You need to add the docusign_private_key_demo file to the config/keys directory for this to work (django will error when the file is tried to be accessed if you forget it)
2. Right now we are only using the developer version of docusign so none of the documents are binding. When we want to start paying we can update the keys, account etc and then the documents become llegally binding. Though the functionality of the developer version is identical except that the document says for demonstration purposes only on it


config/wsgi.py
```diff
@@ -14,10 +14,16 @@
...
 from config.settings.base import BASE_DIR
 
+from config.keys.keys_filepaths import KEY_FILE_PATHS
+
 # Opens the secrets file and loads the values
 with open(os.path.join(BASE_DIR, 'settings/secrets.json')) as f:
         secrets = json.loads(f.read())
 
+# Tests loading all the keys to make sure they exists (does not verify that they are valid)
+for key, path in KEY_FILE_PATHS.items():
```
tiusty, line 24: @akashkumar2296 + @seanrayment I was trying to think of a way to ensure that all the necessary private keys are in the correct spot before the website starts working. I.e it fails immediately if one is missing (instead of for docusign it only fails when you try to use docusign, so it might be missed). This is my "hack" aka in wsgi try to see if all the private keys exist and therefore all the private key paths are stored in the keys directory as well as with the file that stores the path. Remember that the keys are not going to be stored on VC for security concerns. Let me know if you think this is a bad idea or if you can think of a better way.

cocoon/survey/views.py
```diff
@@ -431,3 +447,120 @@ def delete_visit_house(request):
...
+                                )
+    else:
+        return HttpResponse(json.dumps({"result": "0",
+                                        "message": "Method Not POST"}),
+                            content_type="application/json",
+                            )
+
+
+@login_required
+def resend_pre_tour_documents(request):
```
seanrayment, line 518: Does it make sense to impose a limit on this request? A bad actor could spam this request and we'd quickly hit our docusign request limit/get charged a lot of money

config/wsgi.py
seanrayment: This seems totally reasonable to me

seanrayment (approved):
This is a pretty big PR so I didn't scrutinize every line (esp. because a lot was interacting with docusign API) but things were looking good while toying around locally. 

I left one comment for you to consider, but otherwise LGTM

cocoon/survey/views.py
tiusty: yes I was thinking about that.... I really should. Since the functionality of when it gets called will have to be reimplemented later, I figured I would it then. This PR mostly adds the API interface so we can easily interface with django, though the functionality of how we are going to use it will change a lot.

PR #206, opened 2018-10-19
Issue 49 user auth email verification

tiusty:
Adds email authentication when the user creates an account. Now users cannot login/user their account until they verify their account via through their email. This will prevent spam accounts and force users to use a real email address.


seanrayment (approved):
This looks good

akashkumar2296 (approved):

cocoon/scheduler/models.py
```diff
@@ -38,7 +23,7 @@ class ItineraryModel(models.Model):
     itinerary = models.FileField(blank=True)
     agent = models.ForeignKey(MyUser, related_name='scheduled_tours', on_delete=models.SET_NULL, blank=True, null=True)
     tour_duration_seconds = models.IntegerField(default=0)
-    selected_start_time = models.OneToOneField('TimeModel', on_delete=models.SET_NULL, blank=True, null=True)
+    selected_start_time = models.DateTimeField(default=timezone.now, blank=True, null=True)
```
seanrayment: I think we want this to default to `None`

seanrayment (approved):
One comment but otherwise looks ready to merge to me

cocoon/scheduler/models.py
tiusty: Lol good catch! yes it should definitely be none lol

PR #207, opened 2018-10-22
Issue 202 move foreign key

seanrayment:
## Overview of changes

This PR removes `ForeignKey(TimeModel)`, replacing it with the `TimeModel` field, `ForeignKey(ItineraryModel)`.

This was necessary because we have **many** TimeModel's to **one** ItineraryModel. 

I also changed a couple `on_delete`s. We should be able to delete the agent and selected start time without cascading that to the itinerary itself (another agent could pick up the tours, or a new start time could be selected)

#### Note:
This type of PR cannot be merged automatically. In the future, we should have a protocol for updating the db schema to minimize downtime, but here I think it should suffice to delete the scheduler migrations and DROP the available_start_times column from `scheduler_itinerarymodel`.

### DIFF
--- a/cocoon/scheduler/admin.py
+++ b/cocoon/scheduler/admin.py
@@ -4,16 +4,16 @@ from .models import ItineraryModel
 # Register your models here.
 
 
-class ItineraryAdmin(admin.ModelAdmin):
-    fieldsets = [
-        ('Client',
-         {'fields': ('client', 'homes')}
-         ),
-        ('Tour',
-         {'fields': ('itinerary', 'tour_duration_seconds', 'available_start_times')}),
-        ('Agent',
-         {'fields': ('agent', 'selected_start_time')})
-    ]
-
-
-admin.site.register(ItineraryModel, ItineraryAdmin)
+# class ItineraryAdmin(admin.ModelAdmin):
+#     fieldsets = [
+#         ('Client',
+#          {'fields': ('client', 'homes')}
+#          ),
+#         ('Tour',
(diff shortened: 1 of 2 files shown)


tiusty:
This looks good to me! I haven't tested it yet myself but tomorrow I should be able to. Has this worked when you tested it locally?

seanrayment:
@tiusty All tests pass, the admin works, and I played around with the models in the shell. Though like I mentioned in the PR I did have do a bit of manual work on the db to get things working.

tiusty:
sweet! I will approve it so you can merge it if it is blocking you. I will play around with it later tonight anyways to see how it works

tiusty:
Also I realized we should start leaving the __init__.py in the migrations folder. Can you do that for the scheduler app? If you want you can add it to the rest of the apps as well. You will also need to define a !ignore in the gitignore file i.e
!migrations/__init__.py

tiusty:
Also the commands needed to manually fix the issue:

In postgres:
sudo -u postgres psql (connect to postgres)
\connect cocoon (connect to cocoon database)

(drop the necessary tables, please verify the naming etc before actually running them i.e \dt)
DROP TABLE scheduler_itinerarymodel_homes CASCADE;
DROP TABLE scheduler_timemodel CASCADE;
DROP TABLE scheduler_itinerarymodel CASCADE;

Rollback the migration that we don't want:
migrate --fake scheduler zero

then makemigrations and migrate again with the new branches model

tiusty:
@akashkumar2296 you will have to follow the instructions above when the new model gets merged to develop

tiusty (approved):

PR #215, opened 2018-10-27
[issue-209] Got rid of the one to one field and the selected time in …

tiusty:
…not a DateTimeField


seanrayment (commented):
~~Should we make an availabe_start_times field in ItineraryModel?~~

Whoops, didn't see the updated TimeModel at the bottom. This looks good!

seanrayment (approved):

PR #216, opened 2018-10-31
Add agent-side functionality for tour scheduling

seanrayment:
## Changes included in this PR
Allows broker accounts to view unclaimed itineraries at `/scheduler/agentScheduler` and 'claim' itineraries. These itineraries then appear at `/scheduler/myTours`, where brokers can confirm tours by selecting from a list of user provided available start times. 

- [x] URL and view for agents to view pool of unclaimed itineraries 
- [x] AJAX and view functions to support claiming itineraries
- [x] URL and view for agents to view claimed and scheduled itineraries
- [x] AJAX and view functions to support selecting itineraries
- [x] Notifies user by email when a broker has confirmed their itinerary
- [x] AJAX and view functions to support cancelling tour (user side)
- [x] Notifies broker by email when a user has cancelled the tour
- [x] Admin support for itineraries


cocoon/scheduler/admin.py
```diff
@@ -9,15 +10,7 @@ class TimeInline(admin.StackedInline):
     extra = 0
 
 class ItineraryAdmin(admin.ModelAdmin):
-    fieldsets = [
-        ('Client',
-         {'fields': ('client', 'homes')}
-         ),
-        ('Tour',
```
tiusty, line 16: We still want the other fields right? You just want to have the homes show up as raw

cocoon/scheduler/models.py
```diff
@@ -44,7 +47,84 @@ def tour_duration_hours(self):
         """
         return self.tour_duration_minutes/60
 
+    @property
+    def is_claimed(self):
```
tiusty, line 51: Can you just finish off the function documentation to indicate the return type of the property and what True and False represent

cocoon/scheduler/models.py
```diff
@@ -44,7 +47,84 @@ def tour_duration_hours(self):
...
+        # send confirmation email to user
+        email.send()
+
+    @transaction.atomic
+    def associate_agent(self, agent):
+        self.agent = agent
+        self.save()
+
+    @transaction.atomic
+    def unschedule_itinerary(self, **kwargs):
```
tiusty, line 99: Also write what the expected kwargs are in the function doc along with what it does and the return value

cocoon/scheduler/urls.py
```diff
@@ -0,0 +1,12 @@
+from django.conf.urls import url
+
+from . import views
+
+app_name = 'scheduler'
+urlpatterns = [
+    url(r'^agentScheduler/', views.agent_scheduler, name="agentScheduler"),
+    url(r'^myTours/', views.view_tours, name="myTours"),
```
tiusty, line 8: can you write a comment and then put all the ajax request below saying "these are ajax request". Kinda like what I do for the survey urls

cocoon/scheduler/views.py
```diff
@@ -1,3 +1,192 @@
...
+# Models
+from cocoon.houseDatabase.models import RentDatabaseModel
+from cocoon.userAuth.models import UserProfile
+from cocoon.scheduler.models import ItineraryModel, TimeModel
+
+# Python Modules
+import json
+
+@login_required()
+def agent_scheduler(request):
```
tiusty, line 17: I know tests are annoying, but I was thinking that we should start testing the view functions.

More specifically can you test that given a list of itineraries, all the unscheduled itineraries show up? Also test that the itineraries that are scheduled are not in that list. Also test that only broker and admins can see the page

You can either test the view function directly or create a helper module of agent_scheduler functions and just test out that part of the function.

Right now this function is straightforward enough that we may not need tests, but if we have them, then as we add functionality etc, then those tests will become important to verify that we didn't break the page

cocoon/scheduler/views.py
```diff
@@ -1,3 +1,192 @@
...
+            .exclude(agent=None)
+        context['unclaimed_itineraries'] = unclaimed_itineraries
+        context['claimed_itineraries'] = claimed_itineraries
+
+    else:
+        raise Http404("This page does not exist")
+    return render(request, 'scheduler/itineraryPicker.html', context)
+
+@login_required()
+def view_tours(request):
```
tiusty, line 33: same for above comment.

The tests shouldn't take too long to test but will improve the robustness of the code down the line when we add more funcitonality

cocoon/scheduler/views.py
```diff
@@ -1,3 +1,192 @@
...
+        scheduled_itineraries = ItineraryModel.objects.filter(agent=current_profile.user).exclude(selected_start_time=None)
+        context['unscheduled_itineraries'] = unscheduled_itineraries
+        context['scheduled_itineraries'] = scheduled_itineraries
+    else:
+        raise Http404("This page does not exist")
+    return render(request, 'scheduler/viewTours.html', context)
+
+@login_required
+def get_user_itineraries(request):
+    """
```
tiusty, line 47: I think this one is important to test as well.

cocoon/scheduler/views.py
```diff
@@ -1,3 +1,192 @@
...
+            'unscheduled_itineraries': None,
+        }
+
+
+########################################
+# AJAX Request Handlers
+########################################
+
+@login_required
+def unschedule_itinerary(request):
```
tiusty, line 74: ajax functions would be nice to test too if they aren't too difficult

tiusty (changes requested):

cocoon/scheduler/admin.py
seanrayment: It should still be showing all the fields by default. Let me know if you are seeing something else

seanrayment:
@tiusty I've added tests if you want to take another look

cocoon/scheduler/admin.py
tiusty: yea they are... my bad

cocoon/scheduler/tests.py
```diff
@@ -1,3 +1,172 @@
...
+from django.urls import reverse
+from django.contrib.auth import login, get_user
+from django.utils import timezone
 
-# Create your tests here.
+from cocoon.userAuth.models import MyUser, MyUserManager
+from cocoon.scheduler.models import TimeModel, ItineraryModel
+from cocoon.scheduler import views
+
+def setup_test_itineraries(num_claimed=2, num_unclaimed=2):
```
tiusty: minor: can you put these inside the class as member methods? That way there is no chance someone uses them/ they are linked to the Test Class

tiusty (approved):
Just that one minor change and it should be good to go!

cocoon/userAuth/models.py
```diff
@@ -24,6 +24,7 @@ def _create_user(self, email, password, is_superuser, is_admin, **extra_fields):
         if not email:
             raise ValueError('The given email must be set')
         email = self.normalize_email(email)
+        print(**extra_fields)
```
tiusty: Can you remove this print?

tiusty (approved):

cocoon/survey/views.py
```diff
@@ -236,9 +236,9 @@ def form_valid(self, form):
 
 @login_required
 def visit_list(request):
-    context = {
-        'error_message': []
-    }
+
+    context = views.get_user_itineraries(request)
```
tiusty, line 240: @seanrayment This is broken. Could you make a PR hotfix branch to fix this issue on develop?

PR #218, opened 2018-10-31
Client scheduler pr branch

akashkumar2296:


akashkumar2296:
When clicking "Schedule Tours" button, the algorithm is run using user_profile.favorites.all()

This doesn't necessarily hold the scheduler to 1-6 homes but all the favorites of the user. We should change the functionality so that theres a way for the user to only favorite 1-6 before clicking "Schedule tours", or we should run the algorithm separately from the visitList.html page. But as far as I saw, theres no way to change the visit list so the algorithm is being run on all favorites

cocoon/scheduler/clientScheduler/test_client_scheduler_algorithm.py
```diff
@@ -0,0 +1,42 @@
+from django.test import TestCase
+from unittest.mock import MagicMock
+from client_scheduler import clientSchedulerAlgorithm
+
+
+class TestClientSchedulerAlgorithm(TestCase):
```
tiusty: I am not convinced that the alogrithm works at all cases by seeing these tests.
Tests for if the number of nodes is less than 5 would be good, i.e 1, 2 3 etc. Also tests verifying that is working correctly with more configurations. I.e if 2 homes are closer together then 2 homes far away. Then maybe a group of 3 homes close and a group of 2 homes far etc etc.

cocoon/survey/views.py
```diff
@@ -254,17 +264,46 @@ def survey_result_rent(request, survey_url=""):
...
+    for home in user_profile.favorites.all():
+        homes_list.append(str(home))
+
+    client_scheduler_alg = ClientScheduler()
+    shortest_path = client_scheduler_alg.run_client_scheduler_algorithm(homes_list)
+    interpreted_route = client_scheduler_alg.interpret_algorithm_output(homes_list, shortest_path)
+
+    # Create a string so that it can be passed into ContentFile, which is readable in the FileSystem operation
+    s = ""
+    for item in interpreted_route:
```
tiusty: An imporant part of the route is the time it takes to go between each node. Can you append the edge weight to the itinerary file? Also we need to make sure to add time for each home. So add 20 minutes to ethe estimated to for each home that is stopped at

cocoon/survey/views.py
```diff
@@ -254,17 +264,46 @@ def survey_result_rent(request, survey_url=""):
...
+        s += item
+        s += "\n"
+
+    # Update Itinerary Model
+    # File name is unique based on user and current time (making it impossible to have duplicates)
+    itinerary_model = ItineraryModel(client=request.user, selected_start_time=datetime.now())
```
tiusty: I would make sure that an itinerary model doesn't already exist before creating one. Otherwise a new itinerary model is created on each page load

cocoon/scheduler/clientScheduler/client_scheduler.py
```diff
@@ -0,0 +1,67 @@
+# import distance matrix wrapper
+from cocoon.commutes.distance_matrix.distance_wrapper import DistanceWrapper
+from cocoon.scheduler.clientScheduler.base_algorithm import clientSchedulerAlgorithm
+
+
+class ClientScheduler(object):
+
+    def __init__(self):
+        self.algorithm = clientSchedulerAlgorithm()
```
tiusty: can you have the ClientScheduler inherit the cilentScheduler Algorithm instead of a member data?

cocoon/scheduler/clientScheduler/base_algorithm.py
```diff
@@ -0,0 +1,45 @@
+class clientSchedulerAlgorithm:
+
+    def find_global_min(self, homes_matrix):
+        '''
+        Find the global minimum of the homes_matrix, basically by just finding the smallest point in the matrix and
+        returning its index
+        :param homes_matrix: The matrix calcualted using DistanceWrapper() with distances between every pair of homes in
+        favorited list
+        :return: starting_point: the Global Minimum. This is the smallest possible distance, so the algorithm will start
```
tiusty: can add you add the data type for the arguments and the return type like I did in my function. I.e (int). Do it for the other functions as well

tiusty (changes requested):

.gitignore
```diff
@@ -52,3 +52,7 @@ db.sqlite3
 
 # The log file generated #
 logCocoon.log
+
+#DS store (mac)
```
tiusty: What is this for?

cocoon/scheduler/clientScheduler/client_scheduler.py
```diff
@@ -0,0 +1,73 @@
+# import distance matrix wrapper
+from cocoon.commutes.distance_matrix.distance_wrapper import DistanceWrapper
+from cocoon.scheduler.clientScheduler.base_algorithm import clientSchedulerAlgorithm
+
+
+class ClientScheduler(clientSchedulerAlgorithm):
+
+    def __init__(self):
+        self.wrapper = DistanceWrapper()
+        clientSchedulerAlgorithm.__init__(self)
```
tiusty: this isn't the correct way to call a parents constructor function,
should be:
`super().__init__()`

cocoon/scheduler/clientScheduler/client_scheduler.py
```diff
@@ -0,0 +1,73 @@
...
+        :return: (list): homes_matrix a 2x2 matrix of distances found using the DistanceWrapper() to indicate the distances betweeen any two homes
+        """
+
+        homes_matrix = []
+
+        for home_one in homes_list:
+
+            home_one_distances = []
+
+            result_distance_wrapper = self.wrapper.get_durations_and_distances(origins=[home_one],
```
tiusty: you don't want to call the get_durations_and_distances function, use the new retrieve_exact_commutes instead and pass in driving manually to make it clear what it is doing

cocoon/survey/views.py
```diff
@@ -254,17 +264,56 @@ def survey_result_rent(request, survey_url=""):
...
+        s += "\n"
+
+    # Update Itinerary Model
+    # File name is unique based on user and current time (making it impossible to have duplicates)
+
+    try:
+        itinerary_model = ItineraryModel.objects.get(client=request.user)
+
+    except ItineraryModel.DoesNotExist:
+        itinerary_model = ItineraryModel(client=request.user, selected_start_time=datetime.now())
```
tiusty: we do not want the selected_start_time to be set. The selected start time is set when the agent selects a start time based on all the available start times

cocoon/survey/views.py
```diff
@@ -254,17 +264,56 @@ def survey_result_rent(request, survey_url=""):
...
+
+    itinerary_model.itinerary.save(os.path.basename("itinerary_route_" + str(request.user)) + "_" + str(datetime.now()),
```
tiusty: The file path should not be generated here, it should be generated in the model's file like I did for the MLSPIN requester images. I.e create a helper a callable that generates the file path

cocoon/survey/views.py
```diff
@@ -254,17 +264,56 @@ def survey_result_rent(request, survey_url=""):
...
+                                   ContentFile(s))
+    itinerary_model.save()
```
tiusty: I don't see the homes getting added to the itinerary. The homes that are used to create the route then needs to be saved back into the itinerary so we have a record of what the homes were

cocoon/scheduler/clientScheduler/base_algorithm.py
```diff
@@ -0,0 +1,57 @@
...
+    def find_global_min(self, homes_matrix):
+        '''
+        Find the global minimum of the homes_matrix, basically by just finding the smallest point in the matrix and
+        returning its index
+        :param homes_matrix: The matrix calcualted using DistanceWrapper() with distances between every pair of homes in
+        favorited list
+        :return: (int): starting_point - the Global Minimum. This is the smallest possible distance, so the algorithm will start
+        from this address
+        '''
+        min_time = 1000000000
```
tiusty: a more robust solution would be to set the min_time to house_matrix[0][1], the first non-zero distance

cocoon/survey/views.py
```diff
@@ -254,17 +264,56 @@ def survey_result_rent(request, survey_url=""):
 
 @login_required
 def visit_list(request):
+    """
+        AJAX request used to run the client scheduler algorithm and also render the visitList.html page. This is
```
tiusty, line 273: This comment is not true, this isn't an AJAX request. Anyways when you pull in develop this function will get rewritten anyway.

cocoon/survey/views.py
```diff
@@ -254,17 +264,56 @@ def survey_result_rent(request, survey_url=""):
...
+        the main driver function that uses all methods of the client scheduler algorithm and outputs the itinerary
+        list file to s3 and populates an instance of the itinerary model.
+        :param request: The HTTP request
+        :return: HttpResponse if everything goes well. It returns a lot of context variables.
```
tiusty, line 277: Lol

cocoon/survey/views.py
```diff
@@ -254,17 +264,56 @@ def survey_result_rent(request, survey_url=""):
...
-
     # Retrieve the models
     user_profile = get_object_or_404(UserProfile, user=request.user)
     (manager, _) = HunterDocManagerModel.objects.get_or_create(
         user=user_profile.user,
     )
 
-    # Since the page is loading, update all the signed documents to see if the status has changed
+    # Run the client scheduler algorithm
+    homes_list = []
```
tiusty, line 293: Views are not meant to have a lot of functionality. I updated the renting survey/survey result to reflect that. If you could put this into a def run() function in the client scheduler that would be great. Then you only pass in a few arguments and it does it within the client_scheduler module instead of in the view.

tiusty (changes requested):

.gitignore
akashkumar2296: Its part of the mac OS, .DS_Store should always be gitignored

cocoon/scheduler/clientScheduler/client_scheduler.py
```diff
@@ -0,0 +1,114 @@
...
+from cocoon.scheduler.clientScheduler.base_algorithm import clientSchedulerAlgorithm
+from cocoon.scheduler.models import ItineraryModel, itinerary_directory_path
+
+# import django modules
+from django.core.files.base import ContentFile
+
+
+class ClientScheduler(clientSchedulerAlgorithm):
+
+    def __init__(self):
```
tiusty: You actually can just delete the __init__() function since it can just inherit it from the base class. It isn't doing anything anymore.

cocoon/scheduler/models.py
```diff
@@ -135,7 +138,9 @@ def unschedule_itinerary(self, **kwargs):
         # send confirmation email to user
         email.send()
         
-        
+def itinerary_directory_path(user):
```
tiusty: Actually I meant that it should be a callable for the file_field I can show you at the meeting

cocoon/scheduler/clientScheduler/tests/test_client_scheduler_wrapper.py
```diff
@@ -0,0 +1,156 @@
...
+        interpreted_route = self.clientScheduler.interpret_algorithm_output(homes_list, shortest_path)
+        self.assertEqual(interpreted_route, [("home3",0), ("home2",2), ("home1",1)])
+
+    def test_interpret_algorithm_output_random(self):
+        '''
+        Straight forward interpretation, should be same as input
+        :return:
+        '''
+
+        homes_list = ["home1", "home2", "home3"]
```
tiusty, line 153: Don't you actually want to test with real homes since that is what the function is actually going to be called with? Also add the assertcalledwith functions to assert that the retrieve exact commutes are called properly with the right arguments

tiusty (changes requested):

tiusty (approved):

PR #222, opened 2018-11-05
Issue #220 Add React

nel-co:
1. Adds webpack to bundle React components into `frontend/static/frontend/js/bundle.js`
2. Pass the React component/props through the template `views.py`:
```
context['component'] = 'survey'
context['props'] = 'prop to pass down'
```
3. Inside `templateName.html`, include `react_base` into the `endScripts` block
```
{% block endScripts %}
    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
    <script src="{% static 'bootstrap-3.3.7-dist/js/bootstrap.js' %}"></script>
    <script src="{% static 'Unicorn/main/navbar.js' %}"></script>
    {% include 'frontend/react_base.html' %} <---- Add this line ----->
{% endblock %}
```
4. After you create a new component, `import` into `frontend/src/index.js` and it to the pages `object`
```
import React from 'react';
import ReactDOM from 'react-dom';

import Survey from './pages/survey/index'; <---- Import Component ---->

const pages = {
    'survey': Survey <---- Add Component Here ---->
};

// This will render the component based on what you pass from `views.py`
ReactDOM.render(
    React.createElement(pages[window.component], window.props),
    window.react_mount
);
```


.gitignore
```diff
@@ -52,3 +52,5 @@ db.sqlite3
 
 # The log file generated #
 logCocoon.log
+
```
tiusty, line 55: Just add a quick note to explain what the git ignore is for

cocoon/frontend/src/index.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import Survey from './pages/survey/index';
+
+const pages = {
+    'survey': Survey
```
tiusty: Are we sure we want the index.js to import the survey component? This won't extend to other apps. I feel like we don't want an index.js for all the apps and instead have an index.js for each app kinda like what you did  below. Therefore, should we delete this file?

cocoon/frontend/src/pages/survey/index.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
tiusty: I feel like we shouldn't have the directory structure pages/survey/index.js. I feel like just src/survey/index.js is fine. Each app will have its own directory in src. Let me know if you disagree

cocoon/static/js/base.js
```diff
@@ -0,0 +1,5 @@
+$(document).ready(function() {
+    setTimeout(function() {
+        $(".message_timeout").fadeOut(1500);
```
tiusty: wait did you change this? I feel like I added this a while ago

cocoon/survey/templates/survey/reactRentingSurvey.html
```diff
@@ -0,0 +1,103 @@
+{% load static %}
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
```
tiusty: We don't want to duplicate code. You should use the django inheritance (one of the reasons why we are doing a hybrid approach) so you can just inherit the nav bar etc, kinda like what the rentingSurvey.html file does

cocoon/survey/templates/survey/rentingSurvey.html
```diff
@@ -199,6 +199,8 @@ <h4 class="panel-title">
             </div>
         </div>
     </div>
+
+    <div id="survey"></div>
```
tiusty: What does this change do? Is this where the react component going to be loaded?

cocoon/survey/views.py
```diff
@@ -99,7 +99,10 @@ def renting_survey(request):
     context['form'] = form
     context['form_destination'] = destination_form_set
     context['number_of_destinations'] = number_of_destinations
-    return render(request, 'survey/rentingSurvey.html', context)
+
+    context['component'] = 'survey'
+    context['props'] = 'test'
```
tiusty: So I think for now we should leave the old survey page untouched and make a new url for the react survey page. I will make this change tomorrow. Therefore we don't break the old survey page while this page is under development.

webpack.config.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: /\.js$/,
+        exclude: /node_modules/,
```
tiusty, line 16: Also quick comments describing what each part is doing? I know some of them are trivial but it will make it easier to read especially for someone not as familiar with react

tiusty (changes requested):

tiusty:
right now the bundle.js file is included twice (figured one is the old one so you should delete it) but for the other one, I feel like we should not keep it in VC because it will constantly change whenever the react code changes. I know we don't want to download npm and have to compile it on the production machine so we do want to store the production bundle.js but def not the development one. For now can you add the bundle.js to gitingore and then we will try to come up with a good solution for storing the production bundle.js in the VC since it shouldn't change very much

cocoon/survey/views.py
tiusty: So this is easy once my other PR is merged in, so I will merged that in later today and then I will make that change. In the mean time you can work on the other changes!

cocoon/static/js/base.js
nel-co: I had to add it back in since it was missing for some reason. Didn't change anything though.

nel-co:
Just tested, everything's still working. 👍

cocoon/static/js/base.js
tiusty: This file is still in develop, it is in cocoon/static/cocoon/js/base.js so could you remove this one? If there is an issue with importing it then we should figure out why

tiusty (approved):
Once the base.js file is removed, and tested that it still works. It should be good to merge!

cocoon/static/js/base.js
nel-co: No issues with importing. When I first ran webpack while setting it up, it replaced this file in static/js with the bundle so I had to recreate it. All fixed now though.

PR #225, opened 2018-11-07
Issue 223 speed up survey

tiusty:
This speed up the survey a lot!

This is done by instead of doing thousands of database queries ... I just pull the data into memory in one query and then operate on it within python. It also makes the code look much neater.

Also fixes some errors where the rent algorithm was being run twice (i.e on form valid and then on redirect which was also causing issues).


cocoon/commutes/distance_matrix/commute_cache_updater.py
```diff
@@ -82,29 +78,51 @@ def check_all_combinations(self):
...
+            # Generates a queryset of all the child zip_codes that exist for this destination
+            child_zips = ZipCodeChild.objects.filter(base_zip_code=dest_zip). \
+                filter(commute_type=self.destination.commute_type). \
+                values_list('zip_code', 'last_date_updated')
+
+            # Create a python list of zip_codes and their corresponding generation date
+            # i.e list[(zip_code, datetime)]
+            child_zip_codes = []
+            for zip_code, last_update in child_zips:
+                child_zip_codes.append((zip_code, last_update))
```
seanrayment: I tend to prefer the inline notation because it puts the things being added to the list at the beginning of the line and I find it easier to read (and would remove the need for the clarifying comment above)

Also, can we make it a dictionary keyed by the zip code? Below we are checking membership for a list of homes, so a dictionary would make that check constant.
```python
child_zip_codes = {zip_code: last_update for zip_code, last_update in child_zips}
```

cocoon/commutes/distance_matrix/commute_cache_updater.py
```diff
@@ -82,29 +78,51 @@ def check_all_combinations(self):
...
+
+            # For every home, see if the home zip_code matches one of the zip_codes in the child_zip codes
+            #   If not then add it to the failed_list
+            for home in homes:
+
+                # list comprehension to see if the home zip_code exists within the child zip_codes
+                result = [item for item in child_zip_codes if item[0] == home.home.zip_code]
```
seanrayment: If using the dictionary, this could be done in constant time and maybe with a more descriptive variable name?

```python
zip_cached = home.home.zip_code in child_zip_codes
```

cocoon/commutes/distance_matrix/commute_cache_updater.py
```diff
@@ -82,29 +78,51 @@ def check_all_combinations(self):
...
+                    if not (home.home.zip_code, home.home.state) in failed_list:
+                        failed_list.append((home.home.zip_code, home.home.state))
+
+        # If the ZipCodeBase doesn't exist then none of the children exist so add them all to the
+        #   failed list
+        except ZipCodeBase.DoesNotExist:
+            for home in homes:
+                # Make sure there are no duplicate entries
+                if not (home.home.zip_code, home.home.state) in failed_list:
+                    failed_list.append((home.home.zip_code, home.home.state))
```
seanrayment: I know this is an edge case, so worrying about the efficiency might be a bit of premature optimization, but making the `failed_list` a set would make the check unnecessary. i.e.

```python
failed_list.add((home.home.zip_code, home.home.state)) # does nothing if it exists
```

cocoon/survey/cocoon_algorithm/rent_algorithm.py
```diff
@@ -86,24 +88,125 @@ def retrieve_all_approximate_commutes(self):
...
+            # Retrieve all the child zip_codes for the destination commute_type
+            child_zips = ZipCodeChild.objects.filter(base_zip_code=destination_zip). \
+                filter(commute_type=destination.commute_type). \
+                values_list('zip_code', 'commute_time_seconds')
+
+            # Create a python list of zip_codes and their corresponding commute times
+            # i.e list[(zip_code, commute_time_seconds)]
+            child_zip_codes = []
+            for zip_code, distance in child_zips:
+                child_zip_codes.append((zip_code, distance))
```
seanrayment: I think doing this as a dictionary might make sense here as well

```python
child_zip_codes = {zip_code: distance for zip_code, distance in child_zips}
```

cocoon/survey/cocoon_algorithm/rent_algorithm.py
```diff
@@ -86,24 +88,125 @@ def retrieve_all_approximate_commutes(self):
...
+                # list comprehension to see if the home zip_code exists within the child zip_codes
+                result = [item for item in child_zip_codes if item[0] == home.home.zip_code]
+
+                # If there was a match, (there should only be one)
+                if result:
+                    # Store the commute time associated with the zip_code
+                    # Note the list comprehension returns a list of tuples though there should only be one
+                    #   element in the list, thus [0] and then the second element of the tuple is the commute_time
+                    #   in seconds so thus the [1], then to get minutes divide by 60
+                    home.approx_commute_times[destination] = result[0][1]/60
```
seanrayment: Using the above dictionary, this could be

```python
if home.home.zip_code in child_zip_codes:
    home.approx_commute_times[destination] = child_zip_codes[home.home.zip_code] / 60
else:
    home.eliminate_home()
```
This might make it faster from the constant access, but I also think it's more readable than the 1-element list

seanrayment (approved):
Can't wait to get these improvements in. Left a couple notes about optimizations / readability, so let me know what you think. But after going over those, looks good to merge

PR #227, opened 2018-11-16
moved ygl and mls management into 1 model

seanrayment:
Merges the two management models.

Note: as usual, some work will be required to fix the migrations/db after this change.


tiusty:
Made comments on slack. summary was thinking we should move the date field into the Home Provider model directly, then we don't need the extra model still

seanrayment:
Whoops, I misinterpreted what the goal was here. I'll do that instead!

PR #228, opened 2018-11-16
Update static filter and HomeProviderModel

seanrayment:
This PR:

- [x] Updates the survey's static filter to omit homes whose `last_updated` field does not match the corresponding listing provider's `last_updated` field

- [x] Moves the concept of `last_updated` from the now deleted Ygl and Mls management models to the HomeProviderModel

### DIFF
--- a/cocoon/houseDatabase/management/commands/mlspin/MlspinRequester.py
+++ b/cocoon/houseDatabase/management/commands/mlspin/MlspinRequester.py
@@ -7,7 +7,7 @@ import os, sys
 from cocoon.houseDatabase.models import RentDatabaseModel, HomeProviderModel
 from cocoon.houseDatabase.management.commands.mlspin._mls_fields import *
 from config.settings.Global_Config import gmaps_api_key
-from cocoon.houseDatabase.models import HomeTypeModel, MlsManagementModel
+from cocoon.houseDatabase.models import HomeTypeModel
 from cocoon.houseDatabase.constants import MLSpin_URL
 from cocoon.houseDatabase.management.commands.helpers.data_input_normalization import normalize_street_address
 
@@ -220,9 +220,9 @@ class MlspinRequester(object):
                     print("[ Integrity Error ] ")
                     num_integrity_error += 1
 
-        manager = MlsManagementModel.objects.all().first()
-        manager.last_updated_mls = self.update_timestamp
-        manager.save()
+        provider = HomeProviderModel.objects.get(provider="MLSPIN")
+        provider.last_updated = self.update_timestamp
+        provider.save()
 
         print("")
         print("RESULTS:")
--- a/cocoon/survey/cocoon_algorithm/base_algorithm.py
+++ b/cocoon/survey/cocoon_algorithm/base_algorithm.py
@@ -1,8 +1,10 @@
 # Import Python Modules
 from django.utils import timezone
+from django.db.models import F
 
 # Import houseDatabase modules
 from cocoon.houseDatabase.models import RentDatabaseModel
+from cocoon.houseDatabase.models import HomeProviderModel
 
 # Import HomeScore class
 from cocoon.survey.home_data.home_score import HomeScore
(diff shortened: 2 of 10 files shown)


cocoon/houseDatabase/management/commands/mlspin/MlspinRequester.py
```diff
@@ -220,9 +220,9 @@ def parse_idx_feed(self):
                     print("[ Integrity Error ] ")
                     num_integrity_error += 1
 
-        manager = MlsManagementModel.objects.all().first()
-        manager.last_updated_mls = self.update_timestamp
-        manager.save()
+        provider = HomeProviderModel.objects.get(provider="MLSPIN")
```
tiusty, line 223: can you actually switch to using the fields or something like this:


    MLSPIN = 'msl'
    YGL = 'ygl'

    PROVIDER = (
        (MLPSIN, 'Mlspin'),
        (YGL, 'YGL'),
    )

so then we could do .get(provider=Provider.MLSPIN) or something like that? That way the strings aren't hardcoded so if we change it in the future then we don't need to worry about that. 

also we can change the database storage to be 2 characters like the commute type and then the human readable name can be longer

seanrayment: Sure, that works. Though unrelated I am a little unclear on why that query is failing the CI in the survey's static filter. You have any idea?

tiusty: you might also want to do get_or_create so if the models were created then it will create it then

tiusty: this also applies to the YGL requester

tiusty (changes requested):
Looks good, this change might be good to add now rather than later

cocoon/survey/cocoon_algorithm/base_algorithm.py
```diff
@@ -69,6 +71,7 @@ def generate_static_filter_home_list(user_survey):
         house_query =  RentDatabaseModel.objects\
             .filter(price__range=(user_survey.min_price, user_survey.max_price)) \
             .filter(currently_available=True) \
+            .filter(last_updated=HomeProviderModel.objects.get(id=F('provider')).last_updated) \
```
tiusty, line 74: shouldn't this just be

.filter(last_updated=HomeProviderModel.objects.get(provider=survey.provider).last_updated 

or something like that?

tiusty: not sure what the F is doing. Also how does this know which provider to select? i.e either MLS or YGL?

tiusty: wait jk I see what you are doing, you are trying to make sure that all the provider for that home is equal to the last update of the home... need to read into the F more to understand what is happening

tiusty: you should probably still do

.filter(last_updated=HomeProviderModel.objects.get(provider=F('provider')).last_updated)

right?

tiusty: unless you want to do

.filter(last_updated=HomeProviderModel.objects.get(id=F('id')).last_updated)

seanrayment: Hm, from the docs:

>When referencing relational fields such as ForeignKey, F() returns the primary key value rather than a model instance:

~so `F('provider')` should be returning the primary key of the provider associated with some home~

EDIT: It looks like we can compare a field against another field in the same object using the `F` expression, but it will only return the key for foreign keys, not the object itself. This seems like an easy thing to do, but I can't think of a way to get around it.

tiusty: it isn't a relational field though. You are accessing the HomeProvider directly, not through a foreign key. So I think you should try using the field directly

seanrayment: How can we get the provider associated with a home without using the foreign key? We are filtering both MLS and YGL homes

tiusty: oh I see... misread that. Yea we are retrieving it from the home directly so yea it is a foreign key. That seems like it should work? I can take more of a look at it later

PR #229, opened 2018-11-26
Issue 221 add react

tiusty:
Recreates the My Surveys page with react.

It also combines the My Surveys page and the favorites page.

This page also updates the functionality so that homes can be easily favorites/added to visit list and then itineraries can be created. Right now creating the itinerary does nothing once it is created. 

A big thing to watch is to make sure the way I am formatting/structuring React is correct cus we will use some of the same patterns in the future.


akashkumar2296 (approved):

cocoon/survey/views.py
```diff
@@ -313,6 +312,84 @@ def post(self, request, *args, **kwargs):
         return HttpResponseRedirect(reverse('survey:visitList'))
 
 
+class RentSurveyViewSet(mixins.RetrieveModelMixin, mixins.UpdateModelMixin, mixins.ListModelMixin, viewsets.GenericViewSet):
+
+    serializer_class = RentSurveySerializer
```
seanrayment, line 317: Are you using this line?

seanrayment (approved):
Just to wrap my head around this...Is the idea to have a `ViewSet` view handler for each page, where the data that used to be passed through context (and used in the template) is requested asynchronously in react and assigned to the state?

And the method of passing `props` through `context` is just used for the react_base? 

Lastly, is routing still done in Django? It will just re-serve react base which would have been cached by the user's browser?

tiusty:
I should put this in the docs but @nel-co Tell me if I am saying anything wrong

The idea is this:

Each page is routed by Django so it will be reloaded every time the page refreshes or a new page is loaded.

View Function:
Each page will still have its own view function like before except in the simplistic case if you just want to use react and no django rendering then basically the view function will only need to do two things
1. render the correct template
2. Provide the context['component'] which determines which Component is loaded on a page. 
The way this works is the context['component'] is passed to the template. In the react_base template the component context is passed to window.component. Then in the index.js page for the bundle that is loaded the window.component value is passed to the React.render function which determines which Component is loaded. This is because for each bundle multiple pages are supported so we need to tell react which Component to load for that page.

Template:
Each page will also have a template, it basically needs to do a few things.
1. Extend boostrap to get nav bar and formatting
2. have the block for react (we might want to add it to a default template)
 `{% block content %}<div id="reactWrapper"></div>{% endblock %}`  

3. Include the react_base.html
`
{% include 'frontend/react_base.html' %}
`
4. Load the necessary bundle 
`<script src="{% static 'frontend/js/userAuth.bundle.js' %}"></script>` -> Can be switch to any bundle that is required

ViewSets
The viewsets are for DRF so it is just the api. Therefore they do not load any templates etc, they only interact with react via axios calls. This will retrieve/update/create/list etc data for react. So the view set is not attach to serve a single page or app, it is just the api to retrieve data from that app and any page/react component can query it to retrieve data

Hopefully this was helpful

cocoon/survey/views.py
tiusty: Yep! Pretty sure that is called in the react to retrieve all the Rent Surveys.

tiusty: oh you meant the serializer_class line. Yes every viewset must be associated with a serializer_class. As an example, right now since i defined mixins.ListModelMixin, even though didn't explicitly define a fucntion def list(), this is implicit recreated from the mixin and therefore when a get method is sent to the url to this viewset it will generate a list of surveys and return the response to react. How the data is serialized and returned via the api call is determined by the serilalizer_class defined by the viewset... so yes this very needed and actually will fail to run if not included

tiusty: As an note, if you explicit define one of the functionalities defined by the mixins then you can change the return of that method, so you don't necessarily need to use the serializer_class but each viewset still needs the serializer class for the class to compile

seanrayment:
@tiusty, Yep that's all super helpful.

Is there a reason why we don't have a separate entry point (in webpack config) for each component, so we could render exactly what we need from the django template by serving that components unique bundle without specifying it through `window.component`?

tiusty:
I guess we could but then we would have a million different bundles. I guess for now we are just defining a bundle per app and then just specifying which Component in that bundle we want. Not sure what is better

nel-co (approved):
React code looks good!

PR #231, opened 2018-12-02
Issue 181 best transit modes 2

akashkumar2296:
- added traffic option
- traffic option only called when mode is driving, but it can be selected with any transit (nothing happens)
- retrieve_exact_commutes takes in traffic_option
- get_durations_and_distances takes in traffic_option
- unit tests for both functions changed


cocoon/commutes/distance_matrix/distance_wrapper.py
```diff
@@ -121,13 +121,33 @@ def get_durations_and_distances(self, origins, destinations, mode="driving"):
...
+
+            if traffic_option and mode == "driving":
+                # departure time = Tuesday, Dec 4, 2018 4:30pm
+                response_json = distance_matrix.distance_matrix(self.client,
                                                             origin_list[:origin_number],
                                                             destinations[:destination_number],
                                                             units=self.units,
-                                                            mode=mode)
+                                                            mode=mode,
+                                                            departure_time=1543941000,
```
tiusty: wait i thought we said that we were just going to use best guess and then just change the departure time? Otherwise it isn't accurate. Like if they choose use best guess at 4:30pm it will still include the traffic time, just the best guess with traffic vs pessimestic at 4:30pm which will be like the worst possible traffic. If they choose no traffic is should get the commute time at like 4:00am

cocoon/commutes/distance_matrix/commute_retriever.py
```diff
@@ -8,7 +8,8 @@
 from cocoon.commutes.constants import GoogleCommuteNaming
 
 
-def retrieve_exact_commute(origins, destinations, mode):
+def retrieve_exact_commute(origins, destinations, mode=CommuteType.DRIVING, traffic_option=False):
```
tiusty: not sure if I like the traffic_option variable name cus it is kinda confusing given that it is a boolean. Doing with_traffic as the name I feel is more appropriate

cocoon/commutes/distance_matrix/distance_wrapper.py
tiusty: I guess I see what you did with the traffic switch case but I feel like there should just be one call to the google api and then you just adjust the parameters appropriate. That will make the implementation cleaner I feel.

tiusty (changes requested):

tiusty:
Just looking at the bigger picture. How well will this integrate with having transit options? I feel like maybe the retrieve exact commutes should take in a class with all the possible options and then in the get_durations and distance. Options are pulled out that are corresponding to the commute type desired. Otherwise we are going to have a billion different arguments. Though you could delay this to the next PR it might be worth while to set the gorund work so the next PR is easier and cleaner

tiusty:
Also you should def add unit tests for the case when traffic is selected. You can use the past tests as a baseline for your new ones. Though def finish making the changes before doing that. I think you should get rid of the if statement for calling the distance matrix and just parse the args

cocoon/commutes/distance_matrix/distance_wrapper.py
```diff
@@ -122,32 +122,28 @@ def get_durations_and_distances(self, origins, destinations, mode="driving", tra
         origin_number = int(min((100 / destination_number), 25))
 
         # traffic option set to best_guess (default)
+        # departure time = Tuesday, Dec 4, 2018 3:30am - No Traffic
+        departure_time = 1543894200
```
tiusty: Any hardcoded values should be in the constants.py file so can you add the constant value to the constants.py with the comment of what that int represents. then just import the value here. 

You can create variables like:

COMMUTE_TIME_WITH_TRAFFIC
COMMUTE_TIME_WITHOUT_TRAFFIC

or something like that

cocoon/commutes/distance_matrix/distance_wrapper.py
```diff
@@ -122,32 +122,28 @@ def get_durations_and_distances(self, origins, destinations, mode="driving", tra
...
+
         traffic_model_in="best_guess"
```
tiusty: This should be in the constants file. Then you actually don't need to create a temp variable since we don't change it. Instead just put the variable striaght into the distance_matrix function.

I.e
TRAFFIC_MODEL_DEFAULT

or something like that

tiusty (changes requested):
Almost there! It looks good after these changes.

Remember to put the constants variable in the right section and comment what the constant value represents. It should also be in the constants file for that app so in this case it would be commutes/constants.py

cocoon/commutes/constants.py
```diff
@@ -20,3 +20,10 @@ class CommuteAccuracy(Enum):
     DEFAULT = EXACT
 
 
+# Traffic Mode Constants- Note: This needs to be updated every year
+# API does not accept times from the past
+# With traffic = Tuesday Dec 4, 2019 4:30pm
+# Without traffic = Tuesday Dec 4, 2019 3:30am
+COMMUTE_TIME_WITH_TRAFFIC = 1575477036
```
tiusty: lol if this is true this is a bug that needs to be addressed cus this will be forgotten about. Can you just do something like dynamically create the day so maybe just use the timezone function where it takes the current day and grabs the next tuesday and then takes that days time at 4:30pm and 3:30am. Cus we cannot rely on people changing it cus when that time hits next year things will break

tiusty (changes requested):
Critical bug that needs addressing

cocoon/commutes/constants.py
```diff
@@ -26,4 +27,11 @@ class CommuteAccuracy(Enum):
 # Without traffic = Tuesday Dec 4, 2019 3:30am
 COMMUTE_TIME_WITH_TRAFFIC = 1575477036
 COMMUTE_TIME_WITHOUT_TRAFFIC = 1575430236
-TRAFFIC_MODEL_DEFAULT = "best_guess"
\ No newline at end of file
+TRAFFIC_MODEL_DEFAULT = "best_guess"
+
+# Check if your commute time is in the future, if it is then change both times by a increasing it
+# a year, this makes sure that the time is always in the future
```
tiusty: this is kinda hacky... could we just use a datetime object to get the exact time we want instead of trying to switch out time?

This will be  a cleaner solution: https://stackoverflow.com/questions/7274267/print-all-day-dates-between-two-dates/7274316

so basically get the current year and then construct the year with date(year, 12, 4). Then you can convert that to unix time. That way we don't even use unix time in the file so it is much easier to read. It is better to not use unix time for the human interface cus it is hard to read.

tiusty (changes requested):
Last change. I don't like the use of unix time cus it isn't clean

tiusty (approved):

PR #232, opened 2018-12-03
React Agent Scheduler

seanrayment:
Makes the agent scheduler React + DRF friendly

* React component for the agentScheduler page
* Consolidates unclaimed, claimed, and scheduled itineraries into a single page
* Augments the itinerary component to make AJAX requests
* Adds ViewSet's for the unclaimed (marketplace), claimed, and scheduled itineraries


tiusty:
Haven't really looked at the code yet and just tested out the functionality and it looks good! I have a few comments though. I know we said that Nelson will go ahead and design it but he won't have time before we release the beta so Tomas/other agents will probably use your page before Nelson redesign it. Could you add some basic design elements to make it a bit cleaner and easier to use? I.e maybe box off each itinerary, format the homes so they don't look super stretched etc, then it would make it look much cleaner. Also for the claiming itineraries, if the function returns false, i.e the home is already claimed, could you make it so an error message pops up and then maybe disable the claim button? Also when an itinerary is claimed instead of having the button disappear could you add it to the claimed section? I think the easiest way and the way I do it in the favorite homes is to refresh the itineraries, i.e retrieve all of them again like you did in the DidMount. That way we also ensure that the list is updated every time an action is taken. I know this will cause unnecessary requests but I think it is worth it to keep the data more in sync as well. Lastly the itineraries returns Not rendered for client field for itineraries that are not the user so if you could fix that, that would be awesome!

cocoon/frontend/src/scheduler/agentScheduler/agentScheduler.js
```diff
@@ -0,0 +1,144 @@
...
+            })
+
+    }
+
+    renderItinerary = (itinerary, key, showTimes, showClaim) => {
+        return (
+            <Itinerary
+                id={itinerary.id}
+                showTimes={showTimes}
+                showClaim={showClaim}
```
tiusty: Should we have the claim value passed into the Itinerary model? I feel like the itinerary model is used a lot of shouldn't need to know whether it is claimed or not. A button can be rendered below it and determing whether it is claimed or not can be computed based on which list it is in

tiusty: That way it is also easy to update the claimed by just pulling the itineraries for each list again

cocoon/scheduler/urls.py
```diff
@@ -1,13 +1,26 @@
...
+router = routers.DefaultRouter()
+router.register(r'itinerary', views.ItineraryClientViewSet, base_name='itinerary')
+router.register(r'itineraryAgent', views.ItineraryAgentViewSet, base_name='itineraryAgent')
+router.register(r'itineraryMarket', views.ItineraryMarketViewSet, base_name='itineraryMarket')
+
 app_name = 'scheduler'
 urlpatterns = [
-    url(r'^agentScheduler/', views.agent_scheduler, name="agentScheduler"),
-    url(r'^myTours/', views.view_tours, name="myTours"),
+    url(r'^agentSchedulerOld/', views.agent_scheduler, name="agentSchedulerOld"),
```
tiusty: is there any reason to keep the old page?

cocoon/templates/base.html
```diff
@@ -54,6 +54,12 @@
                 <ul class="nav navbar-nav">
                     {% if user.is_authenticated %}
                         {% block nav-mysurvies %}<li><a class="link" href="{% url 'userAuth:surveys' %}">My Surveys</a></li>{% endblock %}
+                        {% if user.is_broker or user.is_admin%}
+                            {% block nav-agent-itinerary %}<li><a class="link" href="{% url 'scheduler:agentScheduler' %}">Itineraries</a></li>{% endblock %}
```
tiusty: I feel like this page should be called like Agent Portal or something more representative

tiusty (changes requested):
Looks good, my biggest comments are in the the block comment I put earlier!

cocoon/templates/base.html
tiusty: I kinda like the name marketplace, so we could just call the tab that

tiusty:
Also if you could add a refresh status button the selected start time that would be awesome. I do the same thing for the documents so you can just copy the same functionality. I.e when the button is clicked then have it switch to saying refresh and remove the onClick so the user cannot spam the backend

tiusty:
now that I think about it I feel like the "market place" and the Agent scheduled and unscheduled itineraries should be on separate pages, otherwise that page contains a lot of functionality and gets cluttered.

The Market place page should show all the available itineraries that are not claimed and then maybe have all the selected times on the side to just remind them when they are busy (though we will not implment that now). On the My Itineraries page then if they are a broker, it will show their scheduled and unscheduled itineraries.

seanrayment:
@tiusty Think I incorporated all the changes you requested!

tiusty (approved):
Address a couple things to get changed but looks good after that!

tiusty:
Just copying and pasting what I mentioned in slack:

The agent scheduler react is looking good. If you can address that one bug by tomorrow then that would be awesome, otherwise just make an issue for it and we can resolve it later. If you could just remove any old files we don't need anymore/any ajax files/urls that would be great. I know we obsoleted any of the urls appended with old, and also remove their unittests. Lastly, if you could remove the template files for those pages and also all the view functions that are no longer called including the ajax calls

cocoon/scheduler/views.py
```diff
@@ -121,13 +170,13 @@ def claim_itinerary(request):
                             return HttpResponse(json.dumps({"result": "1"}),
```
tiusty, line 172: Wait are these ajax view functions still used? can we removed them plus remove them from the urls?

cocoon/scheduler/urls.py
```diff
@@ -11,10 +11,8 @@
 
 app_name = 'scheduler'
 urlpatterns = [
-    url(r'^agentSchedulerOld/', views.agent_scheduler, name="agentSchedulerOld"),
     url(r'^agentSchedulerPortal/', views.AgentSchedulerPortal.as_view(), name="agentSchedulerPortal"),
     url(r'^agentSchedulerMarketplace/', views.AgentSchedulerMarketplace.as_view(), name="agentSchedulerMarketplace"),
-    url(r'^myToursOld/', views.view_tours, name="myTours"),
     url(r'^clientScheduler', views.ClientScheduler.as_view(), name="clientScheduler"),
 
     # AJAX requests below
```
tiusty, line 18: Also remove all the AJAX requests?

cocoon/scheduler/views.py
seanrayment: They are still used I believe

tiusty: actually? What uses them? I thought it was just for the itinerary pages and we rewrote them in react

PR #233, opened 2018-12-03
[issue-221] Quick fix to make the buttons more spread out

tiusty:
Quick PR to make the buttons cleaner on the My Surveys Page.
Also fixes issue where if there are no surveys the page doesn't display anything. Now it says take a survey


akashkumar2296 (approved):

nel-co (commented):
Should we link to the survey page if there's none here?

tiusty:
Yes good point! I will add that

PR #241, opened 2018-12-09
Sr react hotfix

tiusty:
Some quick hot fixes to some issues

1. Remove unique key warning
2. Now if there are no homes/no start times, it says so
3. Fix issue on marketplace/ issue with hashing function
4. Makes the itinerary more clearly segreated
5. Add correctly updating itineraries in marketplace (i.e forgot to include the hashing to that page)
6. Removes ajax functions in favor of view sets
7. Fix updating itineraries when claiming/scheduling


tiusty:
Jk there is some other things I want to fix but I need to study for my final first so you can review it later.

Namely,
1. When an itinerary has a start time selected, it will go to the scheduled list instead of staying in the unscheduled list
2. Convert the ajax calls to a viewset cus the ajax calls seems to be slow compared to the viewsets

tiusty:
@seanrayment The PR is ready for review!

seanrayment (approved):
This looks great! Thanks for cleaning up those AJAX functions

PR #253, opened 2019-01-04
Cleanup ManagementModels

seanrayment:
### Purpose of PR

We have singleton `MlsManagementModel` and `YglManagementModel` that just store the last time the respective feed was updated. We should remove those models and add the DateTimeField to `HomeProviderModel`.

### Changes made

* Removed models
* Added DateTimeField to `HomeProviderModel`
* Updated save checks]
* Updated tests
* Updated admin

### Screenshots

home_provider_admin


tiusty (approved):
Looks good! I added a few changes namely:
1. Mlsrequester script now as an arguments to specify how many homes you want to pull so you can just pull 15 etc
2. Switched the string formatter to be supported by 3.5

PR #254, opened 2019-01-04
Issue 244 currently available fix

seanrayment:
#### Note:
This PR must be merged after #253 as it has been re-based to include the model changes incorporated in that PR.

### Description

This PR adds another static filter to fix the bug described in #244


cocoon/survey/cocoon_algorithm/base_algorithm.py
```diff
@@ -67,6 +68,7 @@ def generate_static_filter_home_list(user_survey):
         """
         # Query the database
         house_query =  RentDatabaseModel.objects\
+            .filter(last_updated=F('listing_provider__last_updated_feed')) \
```
seanrayment, line 71: This is the relevant line to be reviewed (the rest of the diff is only relevant to #253 )

tiusty: cool, once the other PR is in I will do a last review of this one

seanrayment: Other PR and unittests are in!

cocoon/survey/cocoon_algorithm/cocoon_algorithm_tests/tests_base_algorithm.py
```diff
@@ -61,3 +63,33 @@ def test_generate_static_filter_home_list_hunter(self):
...
+        """
+        # Arrange
+        user = MyUser.objects.create(email="test@email.com", is_hunter=True)
+        home_type = HomeTypeModel.objects.create(home_type='House')
+        mls_provider = HomeProviderModel.objects.create(provider="MLSPIN")
+        ygl_provider = HomeProviderModel.objects.create(provider="YGL")
+        survey = self.create_survey(user.userProfile, num_bedrooms=2, max_price=3000)
+        survey.home_type.add(home_type)
+        mls_provider.last_updated_feed = timezone.now()
+        ygl_provider.last_updated_feed = timezone.now()
```
tiusty: calling timezone.now() independently makes this unittest somewhat undeterministic because there is a slight timedelta which probably will never cause an issue but could so lets avoid it.

maybe do something like:
time_updated =timezone.now()

then pass in time_updated to all the functions and do the time delta on the time_updated function. 

This will make it more deterministic (granted the timezone.now() is in itself undeterministic....) because then we are using the same exact time value instead of slightly off. I know this is even more mitigated by the fact that the home provider only stores the date which truncates off the time anyways but I think it should be easy and worth while to fix.

tiusty (changes requested):

cocoon/survey/cocoon_algorithm/cocoon_algorithm_tests/tests_base_algorithm.py
tiusty: Also I was reading that the setup method should be avoided... could you create a new test class and then you can create the provider model in that function with the time_updated variable as per above. Therefore we know the homes and the home provider model will have the exact same time.

seanrayment: By the setup method do you mean calling `create_home`?

tiusty (approved):
Looks good!

PR #262, opened 2019-01-08
Fix mlspin request encoding

tiusty:
Fixes issue with MLSrequester script which was failing due to an ascii conversion error in the lat and long google service. The idx feed was returning characters that were not ascii (not sure what they represented in unicode either...) so the encode/decode trick will remove any characters that are not supported by ascii


seanrayment (approved):
This seems to be the recommended approach elsewhere as well, so looks good

https://stackoverflow.com/questions/6539881/python-converting-from-iso-8859-1-latin1-to-utf-8

tiusty:
ok cool thanks! yea I found it on stack over flow haha

PR #266, opened 2019-01-10
Issue 184 home data ml

akashkumar2296:
- Finished word scraper for mls
- Updated form to look for all amenities
- Added filtering in cocoon algorithm to find relevant homes


cocoon/houseDatabase/management/commands/tests/tests_word_scraper.py
```diff
@@ -0,0 +1,401 @@
+from cocoon.houseDatabase.management.commands.helpers.word_scraper import WordScraper
+from django.test import TestCase
+
```
tiusty: Since we have the file in the helpers directory, could you actually make the tests directory in the helper directory and then put the tests there?

cocoon/houseDatabase/management/commands/helpers/word_scraper.py
```diff
@@ -0,0 +1,16 @@
+class WordScraper:
+
+    def __init__(self):
```
tiusty: can you add the adding the input description to the constructor?

cocoon/houseDatabase/models.py
```diff
@@ -64,6 +64,7 @@ class HouseLocationInformationModel(UpdateBase, models.Model):
     zip_code = models.CharField(max_length=200)
     latitude = models.DecimalField(max_digits=9, decimal_places=6, default=0)
     longitude = models.DecimalField(max_digits=9, decimal_places=6, default=0)
+    laundromat_nearby = models.BooleanField(default=False)
```
tiusty, line 75: Not sure if I like this here because it is kinda different then what I was going for that class.

Can you make a new class called HouseNearbyAmenitiesModel or something? We are gonna add stuff for nearby things at some point anyways

cocoon/survey/forms.py
```diff
@@ -118,7 +126,7 @@ class Meta:
         fields = '__all__'
 
 
-class ExteriorAmenitiesForm(ModelForm):
+class AmenitiesForm(ModelForm):
```
tiusty: I would try to separate out of the Amenities like the survey model. 

I ideal the form classes would match to the survey so we want an ExteriorAmenitiesForm, InteriorAmentiesForm etc.

tiusty: You also seem to be missing the laundry_nearby survey form field

cocoon/survey/models.py
```diff
@@ -110,20 +112,71 @@ def min_price(self):
...
+    wants_gym = models.BooleanField(default=False)
+    gym_weight = models.IntegerField(default=0)
+    wants_storage = models.BooleanField(default=False)
+    storage_weight = models.IntegerField(default=0)
+
+    wants_laundry_in_unit = models.BooleanField(default=False)
+    wants_furnished = models.BooleanField(default=False)
+    furnished_weight = models.IntegerField(default=0)
+    wants_dogs = models.BooleanField(default=False)
+    number_of_dogs = models.IntegerField(default=0)
```
tiusty: these form fields are duplicated from the Interior Amenities. We want the amenities seperated by type so it doesn't become a huge list for one class. Try to follow what we categorized them in the house database model.

tiusty (changes requested):
The biggest thing is the orgainzation of which fields go with which class. We should try to keep it constant with the survey model and house data model. They won't all be identical but we should try to keep it similar

cocoon/survey/forms.py
akashkumar2296: Thats how I had it at first, but I was running into a meta class error when I tried importing both of them to RentSurveyModel

TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases

I struggled with this pretty hard so I just put them all together to make it work. If you can find a solution to this, let me know, because I looked pretty hard and the only solutions I found we're really hacky.

tiusty: not sure why you would have that error cus I am inheriting other form classes into the SurveyForm already so it should just be identical to what I have. Try again and if you get the same issue show me and I will resolve it

tiusty: was there something different about the class you were doing?

akashkumar2296: Nope, nothing at all. It was identical to the exterior amenities class. For some reason it just would not work

cocoon/survey/models.py
akashkumar2296: ^ Same issue as above. They were originally separated but I had to combine them

cocoon/houseDatabase/management/commands/mlspin/MlspinRequester.py
```diff
@@ -128,12 +128,19 @@ def parse_idx_feed(self):
...
+                #Initialize word scraper
+                word_scraper = WordScraper(cells[REMARKS])
+
                 # Set the HomeBaseModel Fields
                 new_listing.street_address = normalize_street_address("{0} {1}".format(cells[STREET_NO], clean_address))
                 new_listing.city = self.towns[str(cells[TOWN_NUM])]["town"]
                 new_listing.state = self.towns[str(cells[TOWN_NUM])]["state"]
                 new_listing.zip_code = cells[ZIP_CODE]
                 new_listing.price = int(cells[LIST_PRICE])
+                if word_scraper.word_finder(["laundromat","nearby"]):
```
tiusty, line 140: now the proximity of words we shouldn't set the other ones to false. A home can have multiple of the laundry features at once... i.e it can have in building and nearby

tiusty (changes requested):
Also can you remove those other idx feed files which are not necessary

tiusty (approved):

PR #268, opened 2019-01-10
[issue-267] Now the users phone number is added to their account

tiusty:
now the user phone number is saved to the databaes

### DIFF
--- a/cocoon/userAuth/forms.py
+++ b/cocoon/userAuth/forms.py
@@ -3,6 +3,7 @@ from django.db import transaction
 from .models import MyUser, UserProfile
 from django import forms
 from cocoon.signature.models import HunterDocManagerModel
+import re
 
 from .constants import HUNTER_CREATION_KEY, BROKER_CREATION_KEY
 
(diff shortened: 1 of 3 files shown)


cocoon/userAuth/forms.py
```diff
@@ -105,9 +106,37 @@ class BaseRegisterForm(UserCreationForm):
...
     )
 
+    phone_number = forms.CharField(
+        required=False,
+        label="Phone number",
+        widget=forms.TextInput(
+            attrs={
+                'class': 'form-control',
+                'placeholder': 'Please use format: ###-###-####',
+                'pattern': '\d{3}[\-]\d{3}[\-]\d{4}',
```
seanrayment, line 116: Would it be worth it to support multiple formats? e.g. 5085056853, (508)505-6853, etc.

And then in `save` capture the relevant 10 digits from the regular expression?

seanrayment (approved):
One question that I could go either way on, but looks good

cocoon/userAuth/forms.py
tiusty: I was getting annoyed with formatting this so I left it at that... yes we should eventually but I just kept it with what the survey was supporting for formatting to keep it the same...

PR #269, opened 2019-01-11
Issue 256 add map for drawing

tiusty:
Now the user can draw an area if they so desire to filter homes outside of that area


cocoon/survey/cocoon_algorithm/base_algorithm.py
```diff
@@ -56,8 +60,51 @@ def populate_survey_homes(self, user_survey):
...
+            to shapely polygonss
+        :param survey: (RentingSurvey) -> The survey
+        :return: (list(shapely.polygons)) -> The list of polygons that are formatted to shapely
+        """
+        polygons = []
+        for polygon_model in survey.polygons.all():
+            vertices = []
+            for vertices_model in polygon_model.vertices.all():
+                vertices.append((vertices_model.lat, vertices_model.lng))
+            polygon = Polygon(vertices)
```
seanrayment, line 82: I think there is one more layer of error-checking that needs to be done between translating the frontend polygon representation into the shapely one. I was able to draw both 1-vertex and 2-vertex polygons in google maps that caused a `ValueError: A LinearRing must have at least 3 coordinate tuples` on this line. 

Actually, on second thought, it would make sense to check that the polygon will by a valid shapely polygon in the `save_polygons` function so that a user cannot submit a broken survey form.

cocoon/frontend/src/survey/rentForm/general/generalForm.js
```diff
@@ -423,3 +489,77 @@ export default class GeneralForm extends Component {
...
+    <GoogleMap
+        defaultZoom={11}
+        defaultCenter={{lat: 42.3601, lng: -71.0589}}
+        styles={
+            {elementType: 'geometry', stylers: [{color: '#242f3e'}]}
+        }
+        defaultOptions={defaultMapOptions}
+    >
+
+        /* Draws all the polygons stored in the state */
```
seanrayment: Funny enough, since this comment is nested in the JSX, it actually gets rendered in the HTML as text. It's mostly covered by the "Delete All Areas" button, but still visible.

draw_on_map_comment

cocoon/frontend/src/survey/rentForm/main/index.js
```diff
@@ -428,6 +432,54 @@ export default class RentForm extends Component {
...
+        // Removes the polygons from the state since we deleted them
+        this.setState({
+            generalInfo: {
+                ...this.state.generalInfo,
+                polygons: [],
+            }
+        })
+    };
+
+    handleCompletePolygon = (p) => {
```
seanrayment, line 448: It might be nice to also have frontend validation that the polygon is valid in here. i.e.
```
if vertices.length < 3 {
    alert("Selected area must have at least 3 points") // ideally a less intrusive error message xD
    this.deletePolygon()
}
```

cocoon/frontend/src/survey/rentForm/main/index.js
```diff
@@ -428,6 +432,54 @@ export default class RentForm extends Component {
...
+        /**
+         * Adds the polygon to the state when it is completed
+         */
+        let polygons = [...this.state.generalInfo.polygons];
+        let polygon = {};
+        let verticies = [];
```
seanrayment: totally nitpicking but I think the spelling is "vertices"

seanrayment (changes requested):
Looks great on the frontend and storing polygons in the db is awesome. Just found one bug regarding invalid polygons that are allowed on the frontend but break shapely's `Polygon` constructor. After that and maybe a unit test for `save_polygons` should be good to go!

cocoon/survey/cocoon_algorithm/base_algorithm.py
tiusty: Yea good catch!

tiusty: yes I was being lazy... so thanks for calling me out on not unit testing lol

tiusty:
@seanrayment Should be all set!

My only unsure thing was when we do the edit survey... we need to check if the polygons exist already before adding them... since this is not trivial right now I am going to push it off to the new survey results page. My plan is that when the polygon is pulled from the server it has the id stored in the dictionary (and ones created by the page won't) so then I can just match by id of the polygon. I tried to implement a polygon vertices matching thing but it is messy and think it will be better by id. That is unless you can think of a better way

seanrayment (approved):
LGTM!

PR #274, opened 2019-01-17
Updates home tile styling and adds additional props

nel-co:
This PR updates the home tile style to match figma mockups. Adds props `canFavorite` and `canVisit` to determine which buttons to render on the tile. Adds `isLarge` prop to determine how to render the tile. i.e. horizontal or square.


cocoon/frontend/src/common/homeTile/homeTile.js
```diff
@@ -17,40 +17,91 @@ class HomeTile extends Component {
...
+
     renderScore(home) {
         /**
          * Renders the score portion of the home tile
          * @type {string} THe home that is being rendered
          */
 
-        // Toggles whether the home has a empty or full heart depending on favorite status
-        let heart_class = "glyphicon";
+        // Toggles whether the home text depending on favorite status
```
tiusty: This comment is wordly weirdly.

nel-co:
Just realized I need to fix the border-radius on the images when the tile is the larger one.

cocoon/frontend/src/common/homeTile/homeTiles.js
```diff
@@ -76,9 +76,14 @@ export default class HomeTiles extends Component {
             );
         // Loads one home with extra info when it was clicked on
         } else {
+            let home = this.props.homes.find(home => home.id === this.state.home_click_id);
             return (
                 <HomeTileLarge
                     home={this.props.homes.find(home => home.id === this.state.home_click_id)}
```
tiusty: can we replace the home={filter()} with the home you used so it is consistent. This prevents doing the filter twice.

tiusty (changes requested):
Couple minor comments.

The other thing that needs to get changed is that bug via removing home from visit list as we discussed on slack

tiusty (approved):

PR #284, opened 2019-01-23
Issue 283 add info to home tile

tiusty:
Adds the amenities information to the homeTile


nel-co (approved):
Front end looks good!

PR #294, opened 2019-01-30
Issue 276 update survey results page

nel-co:
The survey results page is now handled with react.


tiusty (approved):
Looks good.

There is a bug with the sliders that is resovled in a following PR

PR #300, opened 2019-01-31
Fix Agent Scheduler

seanrayment:
This PR creates `itinerary_agent.js` and modifies the `ItineraryAgent` ViewSet to get the agent scheduler compatible with the current method of storing Times.


tiusty:
I know this is from the last time, but when the agent clicks claim itinerary and an email gets sent to the client, it gets sent in UTC time. We should switch the timezone to EST. I did the conversion in the YGL script so you can copy the same method so it return EST

cocoon/frontend/src/scheduler/itinerary/itinerary_agent.js
```diff
@@ -0,0 +1,269 @@
...
+        tour_duration_seconds: null,
+        start_times: [],
+        refreshing: false,
+    };
+
+    updateItinerary() {
+        this.setState({
+            refreshing: true,
+        });
+        axios.get(scheduler_endpoints['itinerary'] + this.props.id + '/')
```
tiusty, line 37: You can do this, or I can do it if you want but here is a flaw that I introduced. Right now we kinda have a "security hole" because any can view anyways itinerary. I did that because the agent can view an itinerary that is not necessarily their own. What we should do is retrieve the homes/ data from the respective itinerary viewset. So from the client viewset (so they are restricted to their own itinerarys) or if they are an agent, then they can use the agentviewset which then lets them get access to all the itineraries. If we can switch this method here to use the itineraryAgentViewset that would be awesome. There will probably be some code to change.

tiusty: but hopefully not that much

cocoon/frontend/src/scheduler/itinerary/itinerary_agent.js
```diff
@@ -0,0 +1,269 @@
...
+        return valid_start_times;
+    }
+
+    renderSelectTime = (unix_time) => {
+             return (<button onClick={() => {this.selectTimeButton(moment(unix_time).toISOString())}} className="select-time-button">
+                 {this.state.refreshing ? "..." : "select"}
+             </button>);
+    }
+
+    renderSchedulingInformation = () => {
```
tiusty, line 293: another thing that is saved in the itinerary is the itinerary file. Can that be passed back and the link to it rendered with the intinerary. I can then go in a format the file more nicely later. This will tell the agent the order at which the homes were scheulded plus we want to add more info later

cocoon/scheduler/views.py
```diff
@@ -173,17 +175,31 @@ def update(self, request, *args, **kwargs):
...
-
-            # The start time must be one of the available start times for that itinerary
-            try:
-                time = TimeModel.objects.filter(itinerary=itinerary).get(id=time_id)
-                itinerary.select_start_time(time.time)
-                result = True
-            except TimeModel.DoesNotExist:
-                result = False
-                reason = 'Start time is not one of the available start times'
+        print(self.request.data)
```
tiusty: remove the print statement

cocoon/frontend/src/scheduler/itinerary/itinerary_agent.js
```diff
@@ -0,0 +1,269 @@
...
+                                <span className="item-right-text">
+                                    {this.renderSelectTime(unix_time)}
+                                </span>
+                            </div>
+                        );
+                    })}
+                </div>
+            );
+
+        } else if (this.props.viewType === "itineraryAgentScheduled") {
```
tiusty, line 313: not sure how much work this would be, if it isn't too much. Would it be possible to add one more case which lets them set the itinerary to finished? That way once the tour is over they can mark the tour as complete so the user can schedule another tour

cocoon/scheduler/views.py
```diff
@@ -173,17 +175,31 @@ def update(self, request, *args, **kwargs):
...
+                # find an available start time that works
+                qs = TimeModel.objects.filter(itinerary=itinerary)
+                for time_object in qs:
+                    if proposed_time.time + datetime.timedelta(seconds=itinerary.tour_duration_seconds_rounded) <= \
+                                    time_object.time + datetime.timedelta(seconds=time_object.time_available_seconds):
+                        start_time_valid = True
+                        break
+
+                if start_time_valid:
+                    itinerary.select_start_time(proposed_time.time)
```
tiusty, line 197: since you created the propsed_time. Once you set the time, shouldn't you also call propsed_time.delete() since we don't want to keep that timemodel around?

tiusty (changes requested):
Looks good. Few changes

Also my comment about the email which I put above

cocoon/frontend/src/scheduler/itinerary/itinerary_agent.js
tiusty: jk this is a non-issue. I didn't realize I added the condition to prevent this

tiusty (approved):

PR #308, opened 2019-02-01
#133 Restyles the homepage to match figma design

nel-co:


cocoon/homePage/templates/homePage/landingPage.html
```diff
@@ -0,0 +1,252 @@
...
+        <a href="{% url 'scheduler:clientScheduler' %}">My Itinerary</a>
+        <span>/</span>
+        {% if user.is_broker or user.is_admin%}
+          <a href="#">Agent Portal</a>
+          <span>/</span>
+          <a href="#">Marketplace</a>
+          <span>/</span>
+        {% endif %}
+        <div id="dropdown">
+          <button>nelsonc19@gmail.com <i class="material-icons">arrow_drop_down</i></button>
```
tiusty: this is hardcoded, this should be dynamically renderd
i.e {{ user.email }} I am pretty sure

cocoon/homePage/templates/homePage/landingPage.html
```diff
@@ -0,0 +1,252 @@
...
+      <div>
+        <a href="{% url 'userAuth:loginPage' %}">Login</a>
+        <span>/</span>
+        <a href="{% url 'userAuth:hunter_signup' %}" class="redOrange">Sign Up</a>
+      </div>
+
+      {% else %}
+      <!-- Authorized Links -->
+      <div>
+        <a href="{% url 'userAuth:surveys' %}">My Surveys</a>
```
tiusty: we actually only want to render the surveys and itineraries if the user is a hunter
so 
{% if user.is_hunter %}

cocoon/homePage/templates/homePage/landingPage.html
```diff
@@ -0,0 +1,252 @@
...
+            <a href="{% url 'userAuth:user_profile' %}">Profile</a>
+            <a href="{% url 'userAuth:logoutPage' %}">Logout</a>
+          </div>
+        </div>
+      </div>
+        {% endif %}
+    </div>
+
+    <!-- Mobile Nav -->
+    <div class="nav-links_mobile">
```
tiusty, line 66: I like the mobile drop down, but at least on the home page, if you go to the mobile view, the nav bar and the area on the top of the page are the same color, so we should try to fix that... maybe with an outline or something. It looks weird with the brick on brick

tiusty: Maybe we should make that button line more defined so it looks like a different section

cocoon/homePage/templates/homePage/landingPage.html
```diff
@@ -0,0 +1,252 @@
...
+          <p>All rights reserved. Cocoon © 2018</p>
+        </div>
+      </footer>
+
+{% endblock %}
+
+{% intercom_tag %}
+{% block endScripts %}
+    <script src="https://unpkg.com/aos@next/dist/aos.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/tiny-slider/2.9.1/min/tiny-slider.js"></script>
```
tiusty: is there anyway to download this and then store them in with our static files?

tiusty (changes requested):
Overall looks good, some changes that I think are worth while

cocoon/homePage/templates/homePage/landingPage.html
```diff
@@ -0,0 +1,252 @@
...
+
+      {% else %}
+      <!-- Authorized Links -->
+      <div>
+        <a href="{% url 'userAuth:surveys' %}">My Surveys</a>
+        <span>/</span>
+        <a href="{% url 'scheduler:clientScheduler' %}">My Itinerary</a>
+        <span>/</span>
+        {% if user.is_broker or user.is_admin%}
+          <a href="#">Agent Portal</a>
```
tiusty: I realized that these url links are not set up either

nel-co: oh good catch!

tiusty (approved):

tiusty:
Looks good to me!

PR #324, opened 2019-02-05
Release candidate

tiusty:
This is the release candidate for the closed beta release.

There is one change that is missing but everything else should be all set for now


cocoon/frontend/src/common/homeTile/homeTile.css
```diff
@@ -283,4 +283,35 @@
...
+}
+
+
+.itinerary-section-item {
+    display: inline-block;
+    width: 100%;
+    padding: 10px 25px;
+    color: var(--grey);
+    font-weight: 300;
+    border: 1px solid var(--borderColor);
```
nel-co: There's a double border around these where it overlaps. instead of setting the entire `border,` you can just set the `border-top:  1px solid var(--borderColor);` and it would fix it.

To get the alignment right, padding should be `10px 15px`

nel-co (commented):
Looks good! Just had the one comment about the broker info. Not too big of a deal.

cocoon/frontend/src/common/homeTile/homeTile.css
tiusty: thanks, looks much better!

PR #341, opened 2019-02-08
Issue #296 Update verify email look

nel-co:


tiusty:
One thing we should change is the base_email file should not be set for the user authentication. We should remove the stuff within the blocks and instead make that the user authentication email has all those elements. By default the base_html should have nothing inside the block, button etc elements. It should be a blank template

tiusty (changes requested):
Just that one change mentioned above

tiusty (approved):
Looks good!

PR #342, opened 2019-02-09
Issue 310 baseline zipcodes

akashkumar2296:
- Executable: python manage.py baseline_zipcodes
- Output: approximation.txt (JSON object with approximations for all zipcode possibilities)


tiusty:
Looks good a few things:

We should change the script name because it will do more than just baselines i.e
zipcode_handler or something

We need to add an argument to indicate to make the baselines i.e 
python manage.py zipcode_handler --create_baseline

We also want to support transit so we should add another argument to indicate the type
i.e
python manage.py zipcode_handler --create_baseline --commute_type transit
python manage.py zipcode_handler --create_baseline --commute_type driving

the commute type flag will also be used later on for the other functionality.

Also we want to put a check so people don't by accident run the create baseline script:
so if you can add a prompt that says -> "Please don't run this script unless you are directed to, please type yes to confirm"

Can you add all the zipcodes in Massachusetts to the list of zipcodes?
But DON'T run it. I am going to run it at the end of the month after I look at our usage for the month.  I looked it up and there should be like 120 zipcodes in Massachusetts.

Can we rename the output file?
i.e
zipcode_baseline_driving.txt
zipcode_baseline_transit.txt etc

Also it should leave in the directory of the commutes. So can we make sure the output file goes into 
cocoon/houseDatabase/management/commands/baselines/
The easiest way to do it is the import the BASE_DIR variable in the settings.base file. Then you can construct the path above.

Lastly, you can delete the old approximations file because I will run it for all the zip codes later.

Also just to make sure, when you are testing with the distance matrix, try to leave only like 3 zipcodes or something in the list.

tiusty (changes requested):
Based on feedback in comment

PR #349, opened 2019-02-13
Issue 310 baseline zipcodes updates

akashkumar2296:
- Cross references baselines with database and raises PRINT statement (for now) if theres a mismatch in the commute times. TODO: replace with error logging
- Can be done for both commute types
- Creates new instance of zipcode base with random generator for last updated if the instance doesn't already exist in database


cocoon/commutes/distance_matrix/commute_cache_updater.py
```diff
@@ -168,38 +168,41 @@ def generate_approximation_pair(self, origins_zips_states, destination_zip_state
         """
 
         # map (zip, state) tuples list to a list of "state+zip" strings
```
seanrayment, line 170: This comment should be updated since they are `HomeCommute` objects instead of strings now

cocoon/commutes/management/commands/ZipcodeBaseline/ZipcodeBaseline.py
```diff
@@ -0,0 +1,243 @@
...
+                    #   if not then create the pair using the data from the file
+                    if not self.check_key(stored_zipcode_combinations, base_zipcode, child_zipcode):
+                        zip_code_base = ZipCodeBase.objects.get_or_create(zip_code=base_zipcode)[0]
+                        zip_code_base.zipcodechild_set.create(
+                            zip_code=child_zipcode,
+                            commute_distance_meters=data[base_zipcode][child_zipcode][self.JSON_DISTANCE_KEY_NAME],
+                            commute_time_seconds=data[base_zipcode][child_zipcode][self.JSON_DURATION_KEY_NAME],
+                            commute_type=commute_type,
+                            # Spread out the last_date_updated so they don't all fall on the same day
+                            last_date_updated=timezone.now() - timedelta(days=random.randint(0, 14))
```
seanrayment, line 131: Think I'm probably missing something here, but why don't we want to `last_date_updated`s to be the same

cocoon/scheduler/clientScheduler/client_scheduler.py
```diff
@@ -28,23 +28,35 @@ def build_homes_matrix(self, homes_list):
...
+                        home_distances.append(time_seconds)
             else:
-                update_commutes_cache_client_scheduler(homes_list, home_one, accuracy=CommuteAccuracy.APPROXIMATE,
+                update_commutes_cache_client_scheduler(homes_list,
+                                                       home,
+                                                       accuracy=CommuteAccuracy.APPROXIMATE,
                                                        commute_type=CommuteType.DRIVING)
-                home_one_distances = retrieve_approximate_commute_client_scheduler(homes_list, home_one,
-                                                                                   commute_type=CommuteType.DRIVING)
+                home_distances = retrieve_approximate_commute_client_scheduler(homes_list,
```
seanrayment, line 55: I know this isn't a new pattern, but I feel weird about `update_commutes_cache...` and `retrieve_approximate_commute...` being dependent on getting called in that order. Would this break if the call to `update_commutes_cache...` was accidentally left out and the `home_distances` contained -1's?

cocoon/commutes/distance_matrix/distance_wrapper.py
```diff
@@ -101,6 +101,8 @@ def interpret_distance_matrix_response(self, response_obj, mode, with_traffic):
 
                 if origin_distance_list:
                     distance_list.append(origin_distance_list)
+                else:
+                    distance_list.append([(None, None)])
```
seanrayment, line 105: good catch

seanrayment (approved):
Looking good. Just left a few comments

cocoon/commutes/management/commands/ZipcodeBaseline/ZipcodeBaseline.py
tiusty: Because when we load the baselines, we don't want them all the expire at the same time and then need to be updated again. This way this will expire slowly so only a few get updated at a time. This will prevent long loads.

cocoon/scheduler/clientScheduler/client_scheduler.py
tiusty: well it isn't necessary. I guess I should put the update_commute_cache in the retrieve approximate commute funciton tbh. I will do that in a later PR, good catch.

The update is not required to call the retrieve. If a combinations doesn't exist then the retrieve will just return -1 but it won't break the survey. Though with the loaded baselines the combinations should exist.

PR #359, opened 2019-02-19
[issue-248] Now the google maps component doesn't render if the objec…

tiusty:
Now the issues with autocomplete and multiple google api keys are fixed!
The autocomplete should always render now on the survey results page


nel-co (approved):
Looks to be working fine!

PR #378, opened 2019-02-28
Issue 369 gecode commutes backend

tiusty:
Now the geocodes for the destinationes are computed on the backend and loaded to the frontend. Prevents geocoding on every survey load. Also makes it easier to update the commute correctly when it changes on the survey edit


seanrayment (approved):
Looks good 🤘

PR #380, opened 2019-03-02
Individual Itinerary Page

seanrayment:
This PR adds a unique URL for each itinerary for easier viewing. The following changes are included:
- [x] Creates a URL slug for each itinerary by hashing its fields in `post_save`
- [x] Creates a new view, `itineraryPage`, to view an itinerary
- [x] Adds the relevant survey as a field on `ItineraryModel`
- [x] Displays client info (including survey info) and home info on `itineraryFile.html`
- [x] Generates from `listing_agent` numbers to MLSPIN agent pages
- [x] Generates google maps links from addresses

Things missing from this PR to be added later
- [ ] Google map link that finds directions between each home in the correct order
- [ ] Grid view with homes on one axis and amenities on another for easy comparison
- [ ] Adds the time estimates between homes as a field on `ItineraryModel` for viewing on this page
- [ ] Add a button to allow agents to email the itinerary to themself


tiusty (approved):
Added a few changes since I assumed you were probably in bed.
Looks good!

PR #418, opened 2019-03-13
Creates a component to let users see more info about the survey questions.

nel-co:


cocoon/frontend/src/survey/rentForm/main/rentForm.css
```diff
@@ -21,22 +21,13 @@
...
-}
-
-.survey-question h2 span {
-    font-weight: 800;
 }
 
-.checkbox-helper-text {
-    font-size: 12px;
-    color: var(--grey);
+.survey-question .survey-label {
```
tiusty, line 28: Can we put any css for the SurveyQuestion into a css file for the SurveyQuestion component?

cocoon/frontend/src/survey/rentForm/general/generalForm.js
```diff
@@ -309,7 +315,9 @@ export default class GeneralForm extends Component {
     renderPriceWeightQuestion() {
         return (
             <div className="survey-question" onChange={(e) =>this.props.onGeneralInputChange(e, 'number')}>
-                <h2>How <span>important is the price</span>?</h2>
+                <SurveyQuestion
```
tiusty: Maybe we should rename the component SurveyQuestionHeader?

It just contains the header part of the question

cocoon/frontend/src/survey/rentForm/main/rentForm.css
tiusty: That way it is easily resuable and is not coupled into the rentForm.css

cocoon/frontend/src/survey/rentForm/general/generalForm.js
```diff
@@ -170,7 +172,11 @@ export default class GeneralForm extends Component {
     renderNumberOfPeopleQuestion() {
         return (
             <div className="survey-question" onChange={(e) => this.props.onGeneralInputChange(e, 'number')}>
-                <h2>How many people are you <span>searching with</span>?</h2>
+                <SurveyQuestion
+                    surveyQuestion={'How many people are you <span>searching with</span>?'}
+                    hasHelp={true}
+                    surveyQuestionHelpText={'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla mi tortor, pellentesque sit amet ante sed, aliquet varius mi.'}
```
tiusty: One minor thing. Is it easy to add break lines into the argument? That way we can have 2 paragraphs of info if we want. If not no big deal but it might be useful to have

cocoon/frontend/src/survey/rentForm/main/rentForm.css
tiusty: I noticed you did have a survey question css below. Is this not part of the survey question styling then?

tiusty (commented):
Few comments

cocoon/frontend/src/survey/rentForm/main/rentForm.css
nel-co: I'm using `.survey-question` here just to target the `.survey-label`'s that's inside of that div. This is the labels for the radio buttons and checkboxes. I'm changing the component name to SurveyQuestionHeader too, so I think we should just have the styling for the question header inside of the question header css file. What do you think?

tiusty: yes I think we should have all the question heading styling inside of that css file to make it easy to see where the styling is coming from and easy to change in the future

tiusty (approved):
Looks good! I will make another PR afterwards adding the question text to each question

PR #419, opened 2019-03-14
Removes input field from date picker to just use calendar.

nel-co:
Fixes the bug that happens when users type the date into the input field.


tiusty (approved):
Added some changes to get rid of all the extra conversations. Reverted it back to keep the state variable as a moment. (think that is cleaner). Also now on the edit survey, the date picker loads at the months that the user selected

PR #426, opened 2019-03-21
Results page and Map Marker updates

### LINKED ISSUE #400: Go to home pin on survey results
We need a good way of letting the user go to the pin on google map. That is in case the pin is off the map and they need to figure out where it is

nel-co:
This closes #400, #403, and #407 
- Shows more info on the results page map when hovering over home tile/map marker or the home tile is open
- On the results page, when a user clicks the home tile the map marker is centered on the map
- Adds ability to toggle between map/results view on mobile


tiusty (changes requested):
Added some requests in slack

cocoon/frontend/src/survey/resultsPage/map/map.js
```diff
@@ -84,17 +105,78 @@ export default class Map extends Component {
         return mapMarkers;
     };
 
-    render() {
+    createBounds = () => {
```
tiusty, line 117: I am kinda confused what is happening here. Do you mind commenting the code more so it is more clear?

tiusty: Also this seems to just adjust the center lat/lng so how does it set the zoom size?

nel-co: the `createBounds()` was initially going to use the `fitbounds()` method from google maps, but I couldn't get it to work. `fitbounds()` would take the coordinates and center the map around them and set the zoom to a level where they're near the edge.  Since that couldn't happen, `createBounds()` now just sets the center of our map based on the homeTile cooridinates and the zoom level is fixed.

tiusty: ah, I will look into fitbounds now

PR #464, opened 2019-04-14
Issue 445 add video

nel-co:
- Adds a video popup to the landing page
- Removes MailChimp signup section from landing page


tiusty (changes requested):
Added comments on slack.
Looks good but few minor things:
1) Without the maillist the picture is really tall now, it should be chagned back to the original proportions
2) Closing the video before it is done makes the video play in the background with noise, if possible if the div is closed we could mute or stop the video that would be great, though this is minor so if it isn't possible then that is fine

PR #467, opened 2019-04-17
Issue 417 fix word scraper

akashkumar2296:
- fixed function calls
- fixed repretitve words issue


cocoon/houseDatabase/management/commands/helpers/word_scraper.py
```diff
@@ -125,9 +128,70 @@ def look_for_hardwood_floors(self):
...
+        """
+        return self.word_finder(["dishwasher"]) \
+                or self.word_finder(['dish', 'washer'])
+
+    def look_for_pets(self, pet):
+        """
+        Looks for pets using multiple synonyms and abbreviations. Returns True if any one of the
+        conditions match, returns False otherwise
+        """
+        return self.word_finder([pet, "allowed"]) and not self.word_finder(["no",pet, "allowed"])
```
tiusty, line 168: Pretty sure the pet here shouldn't be a variable and should be surrounded by quotation? did that compile??

cocoon/houseDatabase/management/commands/mlspin/MlspinRequester.py
```diff
@@ -137,8 +137,7 @@ def parse_idx_feed(self):
                 new_listing.state = self.towns[str(cells[TOWN_NUM])]["state"]
                 new_listing.zip_code = cells[ZIP_CODE]
                 new_listing.price = int(cells[LIST_PRICE])
-                if word_scraper.word_finder(["laundromat"]):
-                    new_listing.laundromat_nearby = True
+                new_listing.laundromat_nearby = word_scraper.look_for_laundromat()
```
tiusty, line 140: So side note, this script is deprecated and not used anymore.

Now we are using the RETSRequester script instead of the MLS. Could you add the word scrapping there? It is good that we updated this script too in case we ever have to revert to using this. Sorry should have mentioned that

config/keys/keys_filepaths.py
```diff
@@ -4,6 +4,6 @@
 
 # File existence check is done in the wsgi.py file
 KEY_FILE_PATHS = {
-    'DOCUSIGN_PRIVATE_KEY_DEV': os.path.join(BASE_DIR, 'keys', "docusign_private_key_demo.txt"),
+    'DOCUSIGN_PRIVATE_KEY_DEV': os.path.join(BASE_DIR, 'keys', "docusign_private_key_prod.txt"),
```
tiusty: you shouldn't have to and also shouldn't use the production keys, cus this will affect our production environment. Make sure to use the private keys and download the correct file if needed

tiusty (changes requested):
Few changes

.gitignore
```diff
@@ -9,6 +9,9 @@ celerybeat-schedule
 # Ignore the settings file
 config/settings/secrets.json
 
+# Libraries
+cocoon_updated/
```
tiusty: Also this shouldn't be necessarily so can we remove this?

cocoon/houseDatabase/management/commands/helpers/word_scraper.py
akashkumar2296: I realized that we were doing "cats" and "dogs" the same exact way so I made pet a variable in the function signature so that its not repetitive code.

config/keys/keys_filepaths.py
```diff
@@ -4,6 +4,6 @@
 
 # File existence check is done in the wsgi.py file
 KEY_FILE_PATHS = {
-    'DOCUSIGN_PRIVATE_KEY_DEV': os.path.join(BASE_DIR, 'keys', "docusign_private_key_demo.txt"),
-    'DOCUSIGN_PRIVATE_KEY_PROD': os.path.join(BASE_DIR, 'keys', "docusign_private_key_prod.txt"),
+    'DOCUSIGN_PRIVATE_KEY_DEV': os.path.join(BASE_DIR, 'keys', "docusign_private_key_dev.txt"),
```
tiusty: We should just leave these to what they were beforehand. So just revert this change

cocoon/houseDatabase/management/commands/mls_rets/MLSRetsRequester.py
```diff
@@ -118,22 +118,19 @@ def add_home_to_database(self, home):
             word_scraper_remarks = WordScraper(new_listing.remarks)
             word_scraper_appliances = WordScraper(home['Appliances'])
             new_listing.air_conditioning = home['AIR_CONDITION'] == 'Yes'
-            if word_scraper_remarks.look_for_ac() or word_scraper_appliances.look_for_ac():
-                new_listing.air_conditioning = True
+            new_listing.air_conditioning = word_scraper_remarks.look_for_ac()
```
tiusty: well so this actually isn't good because we only want to set the response if it is true. That is because usually the value will set if it in the above line. So this is a second check to set it true if it exists in the word scrapper. This will set it to false if it can't find it in the word scrapper

cocoon/houseDatabase/management/commands/helpers/word_scraper.py
tiusty: oh I missed that

tiusty (changes requested):
Last things

cocoon/houseDatabase/management/commands/mls_rets/MLSRetsRequester.py
tiusty: we actually just want to do what it was doing before, so I would just revert this change as well

tiusty (approved):
Looks good to me!

PR #474, opened 2019-04-19
Itinerary File Additions

seanrayment:
### Additions to the itinerary file:
1. Stores order of home visits in the db
2. Display order of homes on the itinerary file page is now the actual tour order
3. Hides excess home information on itinerary page by default
4. Generates potential viable tour times for each home based on start times, driving distance between homes, and an assumed 20 minute tour time
5. POST functionality for updating whether time slots are still viable
6. Links to a user's survey from the itinerary file


tiusty (approved):
Looks good. At some point we may want to rethink the structure of how we store the itinerary etc to make it cleaner but that can occur in the future and also happen in parallel with making it into react.

